PR #5, opened 2015-03-20
[PATCH V5 0/7] New HP SmartArray Plugin.

cathay4t:
* The new hpsa plugin is based on HP binary tool -- hpssacli which
   could be downloaded from:
    http://downloads.linux.hp.com/SDR/project/spp/

 * This patch set introduced these methods support to hpsa plugin:
    * lsm.Client.systems()
    * lsm.Client.pools()
    * lsm.Client.volumes()
    * lsm.Client.disks()
      # With Disk.STATUS_FREE support also.
    * lsm.Client.volume_raid_info()

 * This patch set only coded and tested on 2 servers(Tony's and mine), a
   automatic tests is running in our internal lab to test this plugin on more
   HP SmartArray cards, will provide subsequent patch if found any bug.

Changes in V2:

 * Patch 2/6:
  * Fix missing '()' in _parse_hpssacli_output().
  * Handle 'No controllers detected' error.

 * Patch 6/6:
  * Fix uninitialized variables in volume_raid_info().
  * Add an extra check to raise PLUGIN_BUG when LD found with no PD found.

 * Please be informed, on RHEL 7.1, you need to manually run 'modprobe sg' 
   command before using this plugin.

Changes in V3:

* Moved 'utils.py' to patch 2/6.

* Patch 2/6:

    Fixed a bug in _parse_hpssacli_output() when 'hpssacli' have this kind of
    message in output:

    ``` 
    Note: Predictive Spare Activation Mode is enabled, physical drives that
    are in predictive failure state will not be available for use as data
    or spare drives.
    ```

    We filter out all line start with 'Note:'

* Patch 4/6: Changed file_read() to use python fileIO(open,read,close).

* Tested on HP SmartArray P420i with plugin_test.py(only got expected failure
  for test_timeout).

Changes in V4:
* Add new patch 7/7: Fix bug of parsing free disks.

Changes in V5:
* Patch 1/7: RPM spec: restart lsmd on plugin package installation.

### DIFF
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -11,3 +11,8 @@ if WITH_MEGARAID
 pluginconf_DATA += pluginconf.d/megaraid.conf
 EXTRA_DIST += pluginconf.d/megaraid.conf
 endif
+
+if WITH_HPSA
+pluginconf_DATA += pluginconf.d/hpsa.conf
+EXTRA_DIST += pluginconf.d/hpsa.conf
+endif
--- a/config/pluginconf.d/hpsa.conf
+++ b/config/pluginconf.d/hpsa.conf
@@ -1,0 +1,1 @@
+require-root-privilege = true;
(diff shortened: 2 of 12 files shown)


tasleson:
One minor nit with not restarting lsmd on install, otherwise I think this is ready.

cathay4t:
@tasleson Thanks. New patch set updated with rpm spec fix.

PR #7, opened 2015-03-30
Bash completion support

tasleson:
* Change default behavior to not do value look ups, override with LSMCLI_AUTO_COMPLETE_VALUE=1

* Added filtering of initiators when an --ag is present on the command line

* Added aliases


cathay4t:
Looks good.

Committed.

PR #11, opened 2015-04-16
[PATCH V6 00/18] Pool RAID info query and hardware RAID volume creation.

cathay4t:
* Putting two patch sets into one. As `pool_member_info()` patch set is
  just preparation for `volume_raid_create()` patch set. It's better to
  explain design choice with both.

* New method `pool_member_info()` to query pool member information:
    * RAID type
    * Member type: disk RAID pool or sub-pool.
    * Members: disks or parent pools.

* New method `volume_raid_create()` to create RAID group volume.

* New method `volume_raid_create_cap_get()` to query RAID group volume create
  supported RAID type and strip sizes.

* Design notes:
    * Why we need pool_member_info() when we already have volume_raid_info()?
        The `pool_member_info()` is focus on RAID disk or sub-pool layout.

        The `volume_raid_info()` is focus on providing information about host
        I/O performance and availability(raid).

    * Why don't create single sharable method for SAN/NAS/DAS about RAID
      creation?
        Please refer to design note of patch 'New feature: Create RAID volume
        on hardware RAID.'

    * Why no RAID creation method for SAN and NAS?
        As we tried before, due to complex RAID settings of SAN and NAS
        storage, I didn't find a simple design to support all of them.

    * Please refer to design note of patch 'New feature: Create RAID volume'
      also.

* Please comment in github page, thank you.
    https://github.com/libstorage/libstoragemgmt/pull/11

Changes in V3:

* Rename pool_raid_info to pool_member_info.
* Rebased patches for the rename.
* Fix C code warning about comparing signed and unsigned integer.
* Initialized pointers in lsm_plugin_ipc.cpp and tester.c.
* Removed unused variables in tester.c.
* volume_create_raid() tested on:
    * MegaRAID: RAID 0/1/5/10.  # No RAID 6 license.
    * HP SmartArray: RAID 0.    # No additional disk.
* Document pull request: https://github.com/libstorage/libstoragemgmt-doc/pull/3

Changes in V4:

* Rename method 'volume_create_raid' to 'volume_raid_create'.
* Fix memory leak in C API.
* Initialize pointers in C code.
* Make sure output pointer is set to sane value if error occurred in C code.
* Add bash auto completion for new lsmcli commands and aliases.
* Fix typo in comments of C and Python library.
* Use hash table to convert vendor RAID string to LSM type in hpsa and
  megaraid plugins.
* Tested MegaRAID for 4 disks RAID 10 with default strip size.
  Query commands is tested on MegaRAID and Smart Array.
* Document updated and C API user guide for these new methods also included:
    https://github.com/libstorage/libstoragemgmt-doc/pull/3

Changes in V5:

* Updated commit message to use current method name:
    * pool_member_info
    * volume_raid_create
    * volume_raid_create_cap_get

Changes in V6:

* Patch 1/18, 8/18:
    Fix memory leaks in C library and C unit test.

* Patch 18/18:
    New patch for fixing volume_raid_create C unit test.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt.h
@@ -864,6 +864,33 @@ int LSM_DLL_EXPORT lsm_volume_raid_info(
     uint32_t *strip_size, uint32_t *disk_count,
     uint32_t *min_io_size, uint32_t *opt_io_size, lsm_flag flags);
 
+/**
+ * Retrieves the membership of given pool. New in version 1.2.
+ * @param[in] c     Valid connection
+ * @param[in] pool  The lsm_pool ptr.
+ * @param[out] raid_type
+ *                  Enum of lsm_volume_raid_type.
+ * @param[out] member_type
+ *                  Enum of lsm_pool_member_type.
+ * @param[out] member_ids
+ *                  The pointer to lsm_string_list pointer.
+ *                  When 'member_type' is LSM_POOL_MEMBER_TYPE_POOL,
+ *                  the 'member_ids' will contain a list of parent Pool
+ *                  IDs.
+ *                  When 'member_type' is LSM_POOL_MEMBER_TYPE_DISK,
+ *                  the 'member_ids' will contain a list of disk IDs.
+ *                  When 'member_type' is LSM_POOL_MEMBER_TYPE_OTHER or
+ *                  LSM_POOL_MEMBER_TYPE_UNKNOWN, the member_ids should
+ *                  be NULL.
+ *                  Need to use lsm_string_list_free() to free this memory.
+ * @param[in] flags         Reserved, set to 0
+ * @return LSM_ERR_OK on success else error reason.
+ */
+int LSM_DLL_EXPORT lsm_pool_member_info(
+    lsm_connect *c, lsm_pool *pool, lsm_volume_raid_type *raid_type,
+    lsm_pool_member_type *member_type, lsm_string_list **member_ids,
+    lsm_flag flags);
+
 #ifdef  __cplusplus
 }
 #endif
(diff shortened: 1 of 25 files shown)


tasleson:
Have you actually been able to test all supported hardware that have more than a couple of disk drives?

cathay4t:
V3 uploaded.

I tested it on MegaRAID for RAID 0/1/5/10 and only tested 1 disk RAID 0 on HP SmartArray.

There is no capability for `volume_create_raid_cap_get()`, 
it's mandatory if `VOLUME_CREATE_RAID` is supported.

cathay4t:
Thanks for the C code review, I will fix those issues as well.

The `volume_raid_create` is OK to me.

Will also use hash look in plugin also.

Through `match.log()` only support log(10), will use log(2) instead.

Thank you again for the kind review.

cathay4t:
Q: Are we certain that all raid implementation will enforce the same limitations when creating a RAID with regards to number of disks?

Tested on MegaRAID for 0/1/5/10. (LSI allows 4 disks RAID 1, but that is actually a RAID 10).
According to HP "Smart Storage Administrator User Guide", they are following the same disk count
restriction.

Meanwhile, as plugin was supposed to  wrap vendor implementation to common LSM interface,
if any vendor does not following LSM RAID disk count restriction, they should be considered as
no support on that raid type. This could save user's time on investigating vendor implementations.

cathay4t:
Q: If we return an error in the second call of values_to_uint32_array, we will leak memory if the user doesn't check the return values. Which raises the question, what should the callers responsibility be in the case where we return an error? In my opinion we should clean up everything so that the caller doesn't have any responsibility to inspect each of the return values in a failing case.

Yes. We should clean everything if `rc != LSM_ERR_OK`.

cathay4t:
V4 patch set uploaded.

cathay4t:
V5 patch set uploaded.

PR #12, opened 2015-04-18
C Example: Fix incorrect constant: LSM_FLAG_RSVD

cathay4t:
* Replace LSM_FLAG_RSVD with LSM_CLIENT_FLAG_RSVD.

 * Add comments about configuration for autotool.


tasleson:
Good catch, thanks!

PR #14, opened 2015-05-01
[PATCH V5 0/4] Volume VPD 83 Fix.

cathay4t:
* After review SPC-3 and SPC-4 document and udev source code, I noticed the
  incorrect definition of our Volume.vpd83.

Changes in V4:

* Patch 4/4: Fix None.lower() error in SMI-S plugin.

Changes in V5:

* Patch 1/4: Remove extra strlen() call by save its return in
  lsm_volume_vpd83_verify() of lsm_mgmt.cpp.

### DIFF
--- a/c_binding/lsm_mgmt.cpp
+++ b/c_binding/lsm_mgmt.cpp
@@ -117,16 +117,23 @@ int lsm_volume_vpd83_verify( const char *vpd83 )
 {
     int rc = LSM_ERR_INVALID_ARGUMENT;
     int i;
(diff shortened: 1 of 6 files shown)


cathay4t:
V5 patch set uploaded.

cathay4t:
Locking this pull request.

I found Linux(udev) is using `ID_WWN_WITH_EXTENSION` instead of `ID_WWN` to
create `/dev/disk/by-id/wwn-*`.
Hence I need to amend the commit message, coments and documents.

PR #34, opened 2015-07-02
MegaRAID Plugin fixes.

cathay4t:
1. Fix incorrect controller ID check in `volume_raid_create()`.
2. Fix pools() method on JBOD or no volume configured.
3. Include user define volume name in volumes() returns.


tasleson:
Looks good Gris.

PR #36, opened 2015-07-03
HP SmartArray plugin: Improve hpssacli output parsing algorithm.

cathay4t:
* The output of hpssacli contains incorrect indented section which
  cause 'duplicate line' PLUGIN_BUG error. To fix it, we only
  parse required section.

* Add more comments to explain the workflow of parsing hpssacli output.

* Tested on SmartArray with and without disk enclosure.

Changes in V2:

* Fix spell errors.

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -74,38 +74,38 @@ def _sys_status_of(hp_ctrl_status):
 def _parse_hpssacli_output(output):
     """
     Got a output string of hpssacli to dictionary(nested).
-    Skipped these line:
-        1. Starts with 'Note:'
-           This is just a message right after controller. We don't neet it
-           yet.
-        2. The 'Physical Drives' line.
-           It should indented after 'Internal Drive Cage' like.
-           If not ignored, we might got duplication line error.
-           After ignored, it's phsycial disks will directly stored as
-           key of 'Internal Drive Cage' dictionary.
+    Workflow:
+    0. Check out top and the second indention level's space count.
+    1. Check current line and next line to determine whether current line is
+       a start of new section.
+    2. Skip all un-required sections and their data and sub-sections.
+    3. If current line is the start of new section, create an empty dictionary
+       where following subsections or data could be stored in.
     """
+    required_sections = ['Array:', 'unassigned']
+
     output_lines = [
         l for l in output.split("\n")
-        if l and not l.startswith('Note:') and
-        not l.strip() == 'Physical Drives']
+        if l and not l.startswith('Note:')]
 
     data = {}
 
-    # Detemine indention level
-    top_indention_level = sorted(
+    # Determine indention level
+    (top_indention_level, second_indention_level) = sorted(
         set(
             len(line) - len(line.lstrip())
-            for line in output_lines))[0]
+            for line in output_lines))[0:2]
 
     indent_2_data = {
         top_indention_level: data
     }
 
+    flag_required_section = False
+
     for line_num in range(len(output_lines)):
         cur_line = output_lines[line_num]
-        if cur_line.strip() == 'None attached':
-            continue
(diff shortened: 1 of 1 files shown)


plugin/hpsa/hpsa.py
```diff
@@ -134,6 +151,7 @@ def _parse_hpssacli_output(output):
             else:
                 cur_data_pointer[cur_line_splitted[0].lstrip()] = \
                     ": ".join(cur_line_splitted[1:]).strip()
+
```
joehandzik: Why does this show up in the diff? Perhaps unintended whitespace?

tasleson: Whitespace, Gris added a blank line.

plugin/hpsa/hpsa.py
```diff
@@ -113,10 +113,27 @@ def _parse_hpssacli_output(output):
...
+        if flag_required_section is False:
+            continue
+
         cur_line_splitted = cur_line.split(": ")
         cur_data_pointer = indent_2_data[cur_indent_count]
 
         if nxt_indent_count > cur_indent_count:
+            # Current line is new section title
             nxt_line_splitted = nxt_line.split(": ")
```
joehandzik: nxt_line_splitted is a rough variable name. nxt_line_split would be better, or even nxt_line_split_res (for next line split result, if you will).

plugin/hpsa/hpsa.py
```diff
@@ -74,38 +74,38 @@ def _sys_status_of(hp_ctrl_status):
...
-        2. The 'Physical Drives' line.
-           It should indented after 'Internal Drive Cage' like.
-           If not ignored, we might got duplication line error.
-           After ignored, it's phsycial disks will directly stored as
-           key of 'Internal Drive Cage' dictionary.
+    Workflow:
+    0. Check out top and the second indention level's space count.
+    1. Check current line and next line to determine whether current line is
+       a start of new section.
+    2. Skip all un-required sections and their data and sub-sections.
```
joehandzik: So, essentially, there's a whitelist of required_sections that breaks your rules about indents? I think that's going to miss a lot of valuable data. As long as this is working for you know, I'm ok with it, but I'll want to writeup an issue to document what this new logic is intentionally not collecting. Make sense?

plugin/hpsa/hpsa.py
joehandzik: Ok, I guess I misinterpreted the diff (it looked like the diff showed that the newline existed already, implying that there was new whitespace "junk" in the line). If it's a clean newline, no concerns on my end.

plugin/hpsa/hpsa.py
cathay4t: @joehandzik The `required_sections` is the only data we care in that plugin right now.
We could increase that variable when we need more information in new methods.
What I did is to save our time to mess with existing[1] and maybe future incorrectly
indented output.

If you have better hpssacli output parse algorithm, please let us know.

[1]: The enclosure and physical disk sections are not indented properly.

plugin/hpsa/hpsa.py
cathay4t: It's an empty new line to separate the loop and final data return.

cathay4t:
Add new patch to fix the 'splitted' error.

cathay4t:
Please ignore the 'The Travis CI build failed ' failure, I haven't set it up yet.
It has been disable for future commit utile I get it done.

plugin/hpsa/hpsa.py
joehandzik: I'll take a look and see what I can accomplish. There's project I'm working on with some of your coworkers at Red Hat that may leverage this library. If we decide to pursue this, I'll definitely work to consume more data.

joehandzik:
Ok, sound fair. Looks good to me!

PR #41, opened 2015-07-22
Travis-CI: Use container for test.

cathay4t:
* Since required packages has been added to whitelist, we are ready to
  use container(sudo: false) way for test.


cathay4t:
The container way test takes 6 min 48 sec(previous was 8 min 14 sec).

tasleson:
Very nice, thanks Gris!

PR #43, opened 2015-07-28
HP SmartArray plugin: Add support for firmware revision and controlle…

joehandzik:
…r mode in lsmcli and python_binding

Signed-Off-By: Joe Handzik <joseph.t.handzik@hp.com>

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1239,6 +1239,8 @@ lsm_system LSM_DLL_EXPORT **lsm_system_record_array_alloc(uint32_t size);
  * @param[in] name          System name (human readable)
  * @param[in] status        Status of the system
  * @param[in] status_info   Additional text for status
+ * @param[in] ctrl_mode     System mode for disk exposure
+ * @param[in] ctrl_fw_ver   System firmware version
  * @param[in] plugin_data   Private plugin data
  * @return  Allocated memory or NULL on error.
  */
@@ -1246,6 +1248,8 @@ lsm_system LSM_DLL_EXPORT *lsm_system_record_alloc(const char *id,
                                                    const char *name,
                                                    uint32_t status,
                                                    const char *status_info,
+                                                   const char *ctrl_mode,
+                                                   const char *ctrl_fw_ver,
                                                    const char *plugin_data);
 
 /**
--- a/c_binding/lsm_convert.cpp
+++ b/c_binding/lsm_convert.cpp
@@ -259,6 +259,8 @@ lsm_system *value_to_system(Value & system)
                                      i["name"].asString().c_str(),
                                      i["status"].asUint32_t(),
                                      i["status_info"].asString().c_str(),
+                                     i["ctrl_mode"].asString().c_str(),
+                                     i["ctrl_fw_ver"].asString().c_str(),
                                      i["plugin_data"].asC_str());
     } else {
         throw ValueException("value_to_system: Not correct type");
@@ -275,6 +277,8 @@ Value system_to_value(lsm_system * system)
         s["name"] = Value(system->name);
         s["status"] = Value(system->status);
         s["status_info"] = Value(system->status_info);
+        s["ctrl_mode"] = Value(system->ctrl_mode);
+        s["ctrl_fw_ver"] = Value(system->ctrl_fw_ver);
         s["plugin_data"] = Value(system->plugin_data);
         return Value(s);
     }
--- a/c_binding/lsm_datatypes.cpp
+++ b/c_binding/lsm_datatypes.cpp
@@ -715,6 +715,8 @@ CREATE_ALLOC_ARRAY_FUNC(lsm_system_record_array_alloc, lsm_system *)
 
 lsm_system *lsm_system_record_alloc(const char *id, const char *name,
                                     uint32_t status, const char *status_info,
+                                    const char *ctrl_mode,
+                                    const char*ctrl_fw_ver,
                                     const char *plugin_data)
 {
     lsm_system *rc = (lsm_system *) calloc(1, sizeof(lsm_system));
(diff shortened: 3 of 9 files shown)


joehandzik:
Hey libstoragemgmt team,

First PR up, this supports the basic data I want to capture for Gen9 controllers:

1. What is the FW version (duh)
2. What mode is the controller in (our Gen9 controllers all support physical disk access, described as HBA mode)

Let me know if I need to add more code elsewhere to finish this one off, I'd also like it if this worked OK for LSI/Avago devices at a minimum. All of the features I'll add in the coming months will be for two reasons:

1. Myself and my team would like to have a better solution for managing HP direct attach storage, and
2. I intend to plug libstoragemgmt into Ceph's RADOS or OSD layer to support storage controller and disk management.

Let me know if you have any questions.

joehandzik:
Looks like I have some unit test work to do...

joehandzik:
Better try rebuilding now. I wasn't clear on if the unit test was failing or if the issue was that I flat-out didn't add support in the c_binding directory for anything I did. I went and added the c_binding code as a first step.

joehandzik:
At this point, I'm just confused about what other nook of this codebase I might be missing. Advice would be appreciated!

joehandzik:
FYI, think I found the nooks, I've just run out of time to fix this today. Surprising how something can work well for you but totally destroy a set of unit tests! DNM unit green, obviously.

cathay4t:
@joehandzik The change you are suggesting will break the version 1.x binary compatibility.
In stead of bumping version to 2.0 at this early stage, try this with full 1.x binary compatibility:
* C
  * `char *lsm_system_fw_version_get(lsm_system *lsm_sys)`
  * `enum lsm_system_hwraid_mode lsm_system_hwraid_mode_get(lsm_system *lsm_sys)`

     This method is dedicate to HW RAID system as SAN/NAS does not need it.
  * `LSM_SYSTEM_HWRAID_MODE_HBA` 

     HBA mode means physical disks are presented to host directly. HW RAID card
     cannot create any RAID group using these physical disks.
  * `LSM_SYSTEM_HWRAID_MODE_RAID`

     RAID mode means only logical(RAIDed) disks are presented to host.
  * New capacity: 
    * `LSM_CAP_SYS_HWRAID_MODE_GET`
    * `LSM_CAP_SYS_FW_VERSION_GET`
* Python
  * Add two constants to lsm.System class
    * `HWRAID_MODE_HBA`
    * `HWRAID_MODE_RAID`
  * Add two methods to lsm.Client
    * `system_hwraid_mode_get(lsm_sys)`
    * `system_fw_version_get(lsm_sys)`
  * Add two constants to lsm.Capability class
    * `SYS_HWRAID_MODE_GET`
    * `SYS_FW_VERSION_GET`

In case you are curious, the Travis CI build failure above is caused by command line test which expecting the `lsmcli ls` returned items only have 4  columns while your patch introduced two 
extra columns.

joehandzik:
Hey Gri,

Thanks for the update, sorry my response has taken so long. I had run into several problems as I tried to get this PR passing all unit tests, your diagnosis sounds right to me. I'll sync up with you outside of Github on some of what I'm trying to do. I'm torn because I understand why you'd want to maintain binary compatibility, but I would like the attributes I'm pulling out to be more than just getters and setters. Does that make sense?

tasleson:
@joehandzik can you please elaborate on "I would like the attributes I'm pulling out to be more than just getters and setters. Does that make sense?

joehandzik:
Yeah, sorry if that was vague. The general instruction is to write a "get_fw" and a "get_controller_state" method, which is fine. It injects an extra call for anyone using this tool though, because one might reasonably expect to only need to make one call to get basic controller metadata instead of two (or more!).

At the surface, that's not really the end of the world and would be a trivial fix for v2 (which is why I've agreed to implement it this way). The broader issue I see with this is that libstoragemgmt (at least for hpssacli) does not cache results, and executes hpssacli every time it wants to look at data, even if a recent query may have retrieved that data. That's not a big deal in low-to-moderate performance environments, but I work with other solution providers who have been able to push enough IO load to make an hpssacli execution cumbersome. Every time you execute hpssacli, the controller forcibly rescans its topology. This involves a number of inquiries going down to each disk. For some storage devices, this equates to a non-NCQ (native command queuing) command, which will have a significant negative impact on IO. 

I'm approaching this from both ends. I'd like the hpssacli team at PMC to implement something that allows an application to avoid such rescans when possible, but I'm also trying to minimize hpssacli executions from the upper layers. Coalescing the "get_fw" and "get_controller_state" methods into the basic "system" method is a good step in that direction.

joehandzik:
Also, sorry for any confusion caused by including "setters" in the discussion. I realize none of what I'll be writing will be setting anything, I just habitually include the terms together. Kinda like peanut butter and jelly.

tasleson:
@joehandzik The function Gris outlined works on an opaque data structure for the system.  There is no round tripping with the get/set attributes on the opaque data types.

Let me outline this a little more.

The client library has a method to retrieve an array of systems, in this case an array of 1 or more HBAs.  The client calls lsm_system_list, the plugin then retrieves all the information (in one command line invocation if possible), creates the system object(s) representing the 1 or more HBAs and then adding the fw version and other added fields incrementally to each system object.

The only performance implication is the function call the client and plugin sides to set/get added field(s).  This is all done in the respective process address space and is in memory.

Yes, this is a little clumsy to have a separate function to retrieve each attribute or set function for each newly added attribute for an opaque type, but it's the only way to incrementally add to a C API and maintain binary compatibility that we are aware of.

joehandzik:
Ok, I suppose I misunderstood the purpose of what Gris had suggested. I didn't realize there would be a way for me to inject the data I'm looking to add into the core lsm_system_list entry. I'll investigate further. My current implementation returns the firmware version and controller state each as separate API calls, so a user of the client library would need to issue separate call for each package of data (core lsm_system data, and the controller's firmware or state data). Thanks!

cathay4t:
@joehandzik Are you still working on this pull requst? If not, let me try from my end with API design I mentioned above.

joehandzik:
The firmware version PR I just created is half of this (as far as I could tell, it followed what you suggested here for the firmware). Give that a look if you can. Also, I should have the other half of this done this week, so no need for you to finish it off unless you already did it.

On Sep 14, 2015, at 7:56 PM, Gri Ge <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> Are you still working on this pull requst? If not, let me try from my end with API design I mentioned above.

�

PR #48, opened 2015-08-05
Bug fix and simplify targetd error paths

tasleson:
Bug fix for volume_unmask and replaced the exception handlers with a common one that maps errors from targetd to error codes with libStorageMgmt with default messages if applicable. This
way we don't need to have a bunch of if/else statements in the exception handlers.

Note: I added a variable so if a caller wants to get the exception to evaluate
they can.

### DIFF
--- a/plugin/targetd/targetd.py
+++ b/plugin/targetd/targetd.py
@@ -16,19 +16,18 @@
 #         Gris Ge <fge@redhat.com>
 
 import copy
-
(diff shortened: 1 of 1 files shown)


tasleson:
The last commit was a trivial one to trigger a CI test which with the updated targetd is now passing.

cathay4t:
@tasleson Besides PEP8 warnings, it looks good to me.

PR #50, opened 2015-09-08
Fix "make check" when running as root.

cathay4t:
Problem:
    "make check" will fail when run it as root with libstoragemgmt user
    account exists.

Root cause:
    lsmd found libstoragemgmt user and drop privilege to that user.
    But the plugin codes are in /root folder where libstoragemgmt user
    has no access, which then lead to failure.

Fix:
    Copy python modules to /tmp/<random>/python folder as we did to
    C lib file.


tasleson:
I was able to duplicate failure and verify change addresses it.

PR #51, opened 2015-09-14
libstoragemgmt: Add support for retrieving firmware versions from storage targets

joehandzik:
Signed-Off-By: Joe Handzik <joseph.t.handzik@hpe.com>

PR #52, opened 2015-09-15
libstoragemgmt: Add support for retrieving hwraid mode from storage t…

joehandzik:
…argets

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt.h
@@ -609,6 +609,19 @@ int LSM_DLL_EXPORT \
                                        char **job, lsm_flag flags);
 
 /**
+ * Retrieves system mode.
+ * @param[in]  conn                  Valid connection
+ * @param[in]  system                lsm_system
+ * @param[out] hwraid_mode           System mode
+ * @param[in]  flags                 Reserved set to zero
+ * @return     LSM_ERR_OK on success, else error reason
+ */
+int LSM_DLL_EXPORT lsm_system_hwraid_mode_get(lsm_connect *conn,
+                                              lsm_system *system,
+                                              uint32_t  *hwraid_mode,
+                                              lsm_flag flags);
+
+/**
  * Retrieves information about the different arrays accessible.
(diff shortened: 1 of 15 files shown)


joehandzik:
We'll see if we hit the same failures. Again, please let me know if there's something structurally wrong with my PRs. My cli code looks...slightly different from yours. But when I try to conform to the model in the rest of the code, I get runtime errors. So let me know if my cli code as it stands is inadequate in particular.

cathay4t:
Please try to use `enum lsm_system_hwraid_mode_type` instead of `uint32_t  *hwraid_mode`.

cathay4t:
Please be informed, once https://github.com/libstorage/libstoragemgmt/pull/51 merged,
this pull request needs the rebase.

joehandzik:
Yup, definitely (re: rebase). Thanks for documenting!

joehandzik:
Hmm, started looking into the enum usage, and I can't say that I agree. If this was a greenfield class I'd implement an enum, but since I put the system mode inside the System python class, I need to make sure the values I have there don't conflict with the corresponding C values for system statuses (see: STATUS_UNKNOWN, STATUS_OK, etc). I then defined values for LSM_SYSTEM_HWRAID states that wouldn't conflict with the other LSM_SYSTEM values. If I create an enum, I wouldn't necessarily be protected from accidentally setting a future HWRAID value equal to an existing LSM_SYSTEM value. 

tldr; I just emulated what already exists for LSM_SYSTEM #defines, and I think it would be more confusing to implement some LSM_SYSTEM values differently from the way existing LSM_SYSTEM values are defined. Is that assertion reasonable, or do you still want me to aggregate LSM_SYSTEM_HWRAID values into an enum?

cathay4t:
@joehandzik To avoid overlap, I am working on this pull request:
1. Use similar design in pull request #56 .
2. Change it to `enum lsm_system_hwraid_mode_type` to `enum lsm_system_mode_type`
    which could expand to include 'SAN, NAS, SAN&NAS, OBJ and etc` values in the future.
3. Add MegaRAID support.

joehandzik:
Sounds good, sorry for the delay @cathay4t.

PR #56, opened 2015-09-22
New API for retrieving system firmware version.

cathay4t:
~~* Please don't commit for these reason, TODO list:
    * Tested on real hardware
    * Check memory leak or etc.
    * Split into small patches.
    * Update license lines.~~
 * Python part:
    * Add new optional argument for `System.__init__()`.
 * C part:
    * ~~Plugin: `lsm_system_record_alloc_v1_3()` with `fw_ver` string.~~
    * Plugin: `lsm_system_fw_version_set()`
    * Client: `lsm_system_fw_version_get()`

 * Misc:
    * Increase version in configure.ac to 1.3.
    * ~~Add empty struct for 1.3 C plugin interface -- `lsm_register_plugin_v1_3()`.~~
    * Fix runtest.sh compatible issue with lsmenv.
    * Included license of Joe@HP Enterprise.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2011-2014 Red Hat, Inc.
+ * Copyright (C) 2011-2015 Red Hat, Inc.
+ * (C) Copyright 2015 Hewlett Packard Enterprise Development LP
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
@@ -14,6 +15,8 @@
  * License along with this library; If not, see <http://www.gnu.org/licenses/>.
  *
  * Author: tasleson
+ *         Joe Handzik <joseph.t.handzik@hpe.com>
+ *         Gris Ge <fge@redhat.com>
  */
 
 #ifndef LIBSTORAGEMGMT_PLUG_INTERFACE_H
@@ -980,8 +983,6 @@ typedef int (*lsm_plug_volume_raid_create) (lsm_plugin_ptr c,
 
 /** \struct lsm_ops_v1_2
  * \brief Functions added in version 1.2
- * NOTE: This structure will change during the developement util version 1.2
- *       released.
  */
 struct lsm_ops_v1_2 {
     lsm_plug_volume_raid_info vol_raid_info;
@@ -1247,6 +1248,18 @@ lsm_system LSM_DLL_EXPORT *lsm_system_record_alloc(const char *id,
                                                    uint32_t status,
                                                    const char *status_info,
                                                    const char *plugin_data);
+
+/**
+ * New in version 1.3. Set firmware version.
+ * @param[in] id            Id
+ * @param[in] sys           System to update.
+ * @param[in] fw_ver        Firmware version string.
+ *                          Caller will get LSM_ERR_INVALID_ARGUMENT for
+ *                          empty string('\0').
+ * @return LSM_ERR_OK or LSM_ERR_NO_MEMORY or LSM_ERR_INVALID_ARGUMENT.
+ */
+int LSM_DLL_EXPORT lsm_system_fw_version_set(lsm_system *sys,
+                                             const char *fw_ver);
 
 /**
  * Retrieve plugin private data
--- a/c_binding/lsm_convert.cpp
+++ b/c_binding/lsm_convert.cpp
@@ -1,5 +1,6 @@
 /*
- * Copyright (C) 2011-2014 Red Hat, Inc.
+ * Copyright (C) 2011-2015 Red Hat, Inc.
+ * (C) Copyright 2015 Hewlett Packard Enterprise Development LP
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
@@ -14,12 +15,17 @@
  * License along with this library; If not, see <http://www.gnu.org/licenses/>.
  *
  * Author: tasleson
+ *         Joe Handzik <joseph.t.handzik@hpe.com>
+ *         Gris Ge <fge@redhat.com>
  */
 
 #include "lsm_convert.hpp"
 #include "libstoragemgmt/libstoragemgmt_accessgroups.h"
 #include "libstoragemgmt/libstoragemgmt_blockrange.h"
 #include "libstoragemgmt/libstoragemgmt_nfsexport.h"
+#include "libstoragemgmt/libstoragemgmt_plug_interface.h"
+
+#define _STD_MAP_HAS_KEY(x, y) (x).find((y)) != (x).end()
 
 bool is_expected_object(Value & obj, std::string class_name)
 {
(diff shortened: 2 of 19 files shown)


c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
```diff
@@ -991,6 +989,14 @@ struct lsm_ops_v1_2 {
     lsm_plug_volume_raid_create vol_create_raid;
 };
 
+/** \struct lsm_ops_v1_2
```
joehandzik: Should be v1_3

c_binding/lsm_convert.cpp
```diff
@@ -255,11 +256,22 @@ lsm_system *value_to_system(Value & system)
     if (is_expected_object(system, CLASS_NAME_SYSTEM)) {
         std::map < std::string, Value > i = system.asObject();
 
-        rc = lsm_system_record_alloc(i["id"].asString().c_str(),
-                                     i["name"].asString().c_str(),
-                                     i["status"].asUint32_t(),
-                                     i["status_info"].asString().c_str(),
-                                     i["plugin_data"].asC_str());
+        if ((i.find("fw_version") != i.end()) &&
```
joehandzik: This is the sort of ugliness in the code that I was afraid of...

c_binding/lsm_plugin_ipc.cpp
```diff
@@ -36,6 +36,7 @@
 #include <libxml/uri.h>
 #include "util/qparams.h"
 #include <syslog.h>
+#include <stdbool.h>
```
joehandzik: I don't see anything that actually uses a boolean, am I missing it?

configure.ac
```diff
@@ -3,7 +3,7 @@ dnl Copyright (C) 2011 Red Hat, Inc.
 dnl See COPYING.LIB for the License of this software
 
 AC_INIT(
-    [libstoragemgmt], [1.2.3], [libstoragemgmt-devel@lists.fedorahosted.org],
+    [libstoragemgmt], [1.3.0], [libstoragemgmt-devel@lists.fedorahosted.org],
```
joehandzik: Ah, now I understand what you wanted me to update. For some reason I was reading the version number as being correct already, sorry!

plugin/sim/simarray.py
```diff
@@ -1725,7 +1725,7 @@ def time_out_get(self, flags=0):
     def _sim_sys_2_lsm(sim_sys):
         return System(
             sim_sys['id'], sim_sys['name'], sim_sys['status'],
-            sim_sys['status_info'])
+            sim_sys['status_info'], '', sim_sys["version"])
```
joehandzik: Did you add two new arguments here, instead of one?

joehandzik:
If this is the preferred style for my more generic commits going forward, I'm fine with that. I'm not the biggest fan of creating version-specific allocators, but you already have version-specific sets of commands...

Personally, I liked your suggestion for creating a branch for each stable release and maintaining binary compatibility within that. If that was done, it'd get rid of some of what you had to do here and I had to do in my other commits. That can be figured out later, I'm more interested in getting the code upstream in some form or another.

plugin/sim/simarray.py
cathay4t: The first one is for `plugin_data`.

c_binding/lsm_plugin_ipc.cpp
cathay4t: No. I forgot to remove that one.

c_binding/lsm_convert.cpp
cathay4t: I will create a macro `MAP_HAS_KEY` for clarity if that was
what you meant.

c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
cathay4t: Thanks. I will fix it.

cathay4t:
Maybe I should try to add a `setter` method for C and python for this new property
in stead of this version-specific allocation.
Coding.

cathay4t:
Tested, splited patches uploaded.
Please review.
Changes:
* Use `lsm_system_fw_version_set()` instead of `lsm_system_record_alloc_v1_3()`.
* Fixed memory leak.
* License lines updated.

c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
```diff
@@ -1249,6 +1250,19 @@ lsm_system LSM_DLL_EXPORT *lsm_system_record_alloc(const char *id,
                                                    const char *plugin_data);
 
 /**
+ * New in version 1.3. Set firmware version.
+ * @param[in] id            Id
+ * @param[in] sys           System to update.
+ * @param[in] fw_ver        Firmware version string.
+ *                          Will got LSM_ERR_INVALID_ARGUMENT for
```
joehandzik: Change line 1257 comment to something like "Caller will get LSM_ERR_INVALID_ARGUMENT..."

c_binding/lsm_convert.cpp
```diff
@@ -260,6 +266,17 @@ lsm_system *value_to_system(Value & system)
                                      i["status"].asUint32_t(),
                                      i["status_info"].asString().c_str(),
                                      i["plugin_data"].asC_str());
+        if ((rc != NULL) && (_STD_MAP_HAS_KEY(i, "fw_version") == 1) &&
```
joehandzik: I actually do like this much better. Nice!

joehandzik:
This is actually a really good compromise between the two version of what I had created and what @tasleson was asking for. Consider it ack-ed by me as soon as you fix the one little typeo I found today.

Now, the main question in my mind is...would you like me to retool #52 in this manner, and store hwraid mode this way?

python_binding/lsm/_data.py
```diff
@@ -357,12 +357,26 @@ class System(IData):
     STATUS_PREDICTIVE_FAILURE = 1 << 4
     STATUS_OTHER = 1 << 5
 
-    def __init__(self, _id, _name, _status, _status_info, _plugin_data=None):
+    def __init__(self, _id, _name, _status, _status_info, _plugin_data=None,
+                 _fw_version=''):
```
joehandzik: As I'm looking at doing something similar in the Disk class...do you need to change the @default_property elements somehow up above?

python_binding/lsm/_data.py
```diff
@@ -357,12 +357,26 @@ class System(IData):
...
         self._status = _status
         self._status_info = _status_info
         self._plugin_data = _plugin_data
+        self._fw_version = _fw_version
+
+    @property
+    def fw_version(self):
+        """
+        String. Firmware version string. New in version 1.3.
+        On some system, it might contains multiple version string.
```
joehandzik: Change to: "On some systems, it might contain multiple version strings."

tools/lsmcli/data_display.py
```diff
@@ -331,6 +331,7 @@ def __init__(self):
     SYSTEM_HEADER['name'] = 'Name'
     SYSTEM_HEADER['status'] = 'Status'
     SYSTEM_HEADER['status_info'] = 'Info'
+    SYSTEM_HEADER['fw_version'] = "FW Ver"
```
joehandzik: Any particular reason for the double quotes?

cathay4t: No. Maybe I should be consistent on using single quote in python.

python_binding/lsm/_data.py
cathay4t: As we need to raise `NO_SUPPORT` error when fw_version is empty, we cannot
depend on `@default_property` decoration.

cathay4t:
@joehandzik Let's wait this acked by Tony before we working on HW RAID mode.

cathay4t:
V2 patch set uploaded.
Changes in V2:
* Fix typos.
* Tested binary compatibilities:
  * Old plugin(simc and sim) <-> new code base.
  * Old client(tester and plugin_test) <-> new code base.

tasleson:
Looks good!  Please update for the 2 things I found too and then we can get this merged.

cathay4t:
V3 patch set uploaded.
Changes in V3:
 * Updated comments of `lsm_system_fw_version_get()` as suggested by Tony.
 * Check input argument by `LSM_IS_SYSTEM()` in `lsm_system_fw_version_get()` and 
   `lsm_system_fw_version_set()`.

tasleson:
Looks good Gris, merged, Thanks!

PR #60, opened 2015-10-14
[PATCH V2 0/8] Adding new system property: system mode.

cathay4t:
Adding new system property: `lsm.System.mode` and `lsm_system_mode_get()`.
Please check patch for detail.

TODO: Need HP's help on hpsa plugin, the original HP patch #52 does not works
            works on my system.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -157,6 +157,8 @@ typedef enum {
 
     LSM_CAP_SYS_FW_VERSION_GET = 160,
     /**^ Plug-in allows user to retrieve storage firmware version */
+    LSM_CAP_SYS_MODE_GET = 161,
+    /**^ Plug-in allows user to retrieve storage mode*/
     LSM_CAP_POOLS_QUICK_SEARCH = 210,
     /**^ Seach occurs on array */
     LSM_CAP_VOLUMES_QUICK_SEARCH = 211,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1262,6 +1262,15 @@ int LSM_DLL_EXPORT lsm_system_fw_version_set(lsm_system *sys,
                                              const char *fw_ver);
 
 /**
+ * New in version 1.3. Set system mode.
+ * @param[in] sys           System to update.
+ * @param[in] mode          System mode 'lsm_system_mode_type'.
+ * @return LSM_ERR_OK or LSM_ERR_INVALID_ARGUMENT.
+ */
+int LSM_DLL_EXPORT lsm_system_mode_set(lsm_system *sys,
+                                       lsm_system_mode_type mode);
+
+/**
  * Retrieve plugin private data
  * @param s     System
  * @return Optional data, NULL if none exist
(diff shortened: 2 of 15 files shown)


cathay4t:
@joehandzik could you check the hpsa plugin patch?
Issues:
1. 'HBA Mode Enabled' does not exist on my system(hpssacli-2.30-6.0.x86_64).
2. If we use 'Controller Mode' instead, please advice the possible value of that property.

joehandzik:
Ah, thanks for pointing that out. I’ll take a look and see what the best alternative is…I’m testing on a Gen9 controller, but it’s good to know that a Gen8 controller (I assume?) yields different output. Best to design around both.

Alternatively, we could do a sets for the attribute, and if it doesn’t exist, the controller is assumed to be in RAID mode.

Joe

cathay4t:
@joehandzik Just FYI. The system I am using is Gen9:
```
[fge@megadeth ~]$ lspci|grep Gen
03:00.0 RAID bus controller: Hewlett-Packard Company Smart Array Gen9 Controllers (rev 01)
[fge@megadeth ~]$ sudo lsmenv hpsa lsmcli ls 
Device alias: hpsa
URI: hpsa://
lsmcli ls
ID             | Name                       | Status | Info | FW Ver | Mode   
------------------------------------------------------------------------------
PDNNF0ARH6W07M | Smart Array P840 in Slot 1 | OK     |      | 2.14   | HW RAID

Time: 0:02.46
```

joehandzik:
That's unexpected. I didn't realize output would be so different between our H series controllers and P series controllers. Thanks for the info!

cathay4t:
* Updated definition to hold LSI MegaRAID JBOD mode[1].
* Updated MegaRAID plugin.


[1]: LSI MegaRAID JBOD mode allows both logical disk and physical disk exposed to OS.

tasleson:
@cathay4t PR looks good, just a few things.  Any updates on SmartArray, your comment indicates we don't have a workable solution at the moment?

joehandzik:
@tasleson Sort of. The implementation I went with initially for Smart Array doesn't work on @cathay4t's Gen9 RAID controller (I was using a Gen9 HBA). Ideally, the hpssacli developers wouldn't have changed the basic templated output between H and P series controllers, but it looks like they did. They have a few attributes that tell you if the controller is in HBA or RAID mode, we'll just need to go with the one that is common between H and P series (which @cathay4t already identified). I'm on the hook for figuring out possible output values...I'll give him my output when in HBA mode, and we'll have enough for an initial implementation between my HBA mode output and his RAID mode output.

joehandzik:
@cathay4t, with an H240, I have:

  Controller Mode: HBA

I'm assuming you have:

  Controller Mode: RAID

?

cathay4t:
@joehandzik Yes. I will change the patch to use "Controller Mode" property. Thanks.

cathay4t:
V2 patch set uploaded.

Changes in V2:

* Fix typo mentioned by Tony. Thanks, Tony.
* Updated SmartArray plugin to use "Controller Mode" to detect system mode.

cathay4t:
@joehandzik Could you please give it a try on your hardware for HBA mode?
I don't have a SmartArray which could[1] configure as HBA mode.
Thank you.


[1]: The server I have is booting from SmartArray, converting to HBA mode requires OS reinstall
      which I would rather not to do so.

joehandzik:
Yup, I'll give it a shot tomorrow!

Joe

On Oct 20, 2015, at 8:26 PM, Gri Ge <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> Could you please give it a try on your hardware for HBA mode?
I don't have a SmartArray which could[1] configure as HBA mode.
Thank you.

[1]: The server I have is booting from SmartArray, converting to HBA mode requires OS reinstall
which I would rather not to do so.

�

cathay4t:
@joehandzik Any update?

joehandzik:
Just tried it, works perfectly (result was the expected value "HBA" for my device)! Sorry for the delay.

cathay4t:
Merged. Thanks.

PR #63, opened 2015-10-26
libstoragemgmt: Add physical disk path information when available

joehandzik:
Here's my first PR. We can talk about what you might want me to change to make this upstreamable here. I can break this down into your more traditional 7-8 separate patches too if you want...

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -157,6 +157,8 @@ typedef enum {
 
     LSM_CAP_SYS_FW_VERSION_GET = 160,
     /**^ Plug-in allows user to retrieve storage firmware version */
+    LSM_CAP_DISK_SD_PATH = 162,
+    /**^ Plugin-in allows user to retrieve scsi device path */
     LSM_CAP_POOLS_QUICK_SEARCH = 210,
     /**^ Seach occurs on array */
     LSM_CAP_VOLUMES_QUICK_SEARCH = 211,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_disk.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_disk.h
@@ -104,6 +104,16 @@ uint64_t LSM_DLL_EXPORT lsm_disk_block_size_get(lsm_disk *d);
 uint64_t LSM_DLL_EXPORT lsm_disk_status_get(lsm_disk *d);
 
 /**
+ * New in version 1.3. Retrieves a disk's scsi device path.
+ * Do not free returned string, free the struct lsm_disk instead.
+ * @param d		Pointer to the disk of interest.
+ * @param sd_path	Pointer to the disk's scsi device path.
+ * @return LSM_ERR_OK on success, else error reason.
+ */
+int LSM_DLL_EXPORT lsm_disk_sd_path_get(lsm_disk *d,
+                                        const char **sd_path);
+
+/**
  * Returns the system id
  * Note: Return value is valid as long as disk pointer is valid.  It gets
  * freed when record is freed.
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1198,6 +1198,15 @@ lsm_disk LSM_DLL_EXPORT *lsm_disk_record_alloc(const char *id,
                                                const char *system_id);
 
 /**
+ * New in version 1.3. Set a disk's scsi device path.
+ * @param disk		Pointer to the disk of interest.
+ * @param sd_path	Pointer to the disk's scsi device path.
+ * @return LSM_ERR_OK on success, else error reason.
+ */
+int LSM_DLL_EXPORT lsm_disk_sd_path_set(lsm_disk *disk,
+                                        const char *sd_path);
+
+/**
  * Allocated the storage needed for one volume record.
  * @param id                    ID
  * @param name                  Name
--- a/c_binding/lsm_convert.cpp
+++ b/c_binding/lsm_convert.cpp
@@ -141,6 +141,18 @@ lsm_disk *value_to_disk(Value & disk)
                                    d["status"].asUint64_t(),
                                    d["system_id"].asString().c_str()
             );
+
+        if ((rc != NULL) && (_STD_MAP_HAS_KEY(d, "disk_sd_path") == 1) &&
+            (d["disk_sd_path"].asC_str()[0] != '\0')) {
+            
+            if (lsm_disk_sd_path_set(rc, d["disk_sd_path"].asC_str()) !=
+                LSM_ERR_OK) {
+                lsm_disk_record_free(rc);
+                rc = NULL;
+                throw ValueException("value_to_disk: failed to update "
+                                     "disk_sd_path");
+            }
+        }
     } else {
         throw ValueException("value_to_disk: Not correct type");
     }
@@ -160,6 +172,8 @@ Value disk_to_value(lsm_disk * disk)
         d["num_of_blocks"] = Value(disk->block_count);
         d["status"] = Value(disk->disk_status);
         d["system_id"] = Value(disk->system_id);
+        if (disk->disk_sd_path != NULL)
+            d["disk_sd_path"] = Value(disk->disk_sd_path);
 
         return Value(d);
     }
--- a/c_binding/lsm_datatypes.cpp
+++ b/c_binding/lsm_datatypes.cpp
@@ -706,6 +706,7 @@ lsm_disk *lsm_disk_record_alloc(const char *id, const char *name,
         rc->block_count = block_count;
         rc->disk_status = disk_status;
         rc->system_id = strdup(system_id);
+        rc->disk_sd_path = NULL;
 
         if (!rc->id || !rc->name || !rc->system_id) {
             lsm_disk_record_free(rc);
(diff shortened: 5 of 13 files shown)


cathay4t:
So this is only for HBA/JBOD mode disk.
How about we set `lsm.Disk.id` as VPD83 NAA ID for HBA/JBOD disk which user could use
`/dev/disk/by-id/wwn-0x<VPD83_NAA>` instead?

joehandzik:
Yeah, this PR is. There's another one that I can push up today that is for /dev/sd* for logical volumes. In fact, that's probably a good idea...push up as many non-interdependent PRs as possible. I could probably have 3-4 (out of 15) outstanding at a given time.

I think the change to lsm.Disk.id would be valuable, and we could use having that symlink available in lsm too, but I NEED /dev/sd*. It's my bootstrap into data exposed by lsscsi, and the data in lsscsi is my bootstrap into being able to light an LED on a drive. Now, one thing I just realized is that we could use the symlink you've proposed to work our way back to /dev/sd* under the covers, and continue to use that as an argument into my lsscsi code...but why bother when the device node comes for free from current hpssacli output?

I think you and I are coming at this from different perspectives. I understand your concerns about this output getting confusing or misleading in more complicated configurations, but that's the case for all the tools we're consuming data from today anyhow (hpssacli would have the same problems that you're concerned about). The reality is that 90+% of the configurations we'll deal with aren't that complicated. I'm all for beefing up the ability to deal with complex configurations, but I don't want that desire to gate features that can be useful immediately.

tasleson:
My initial thought is that a separate retrievable value is preferred.  In general I believe it's better for values to have one meaning only, kind of like 1st normal form form db design.  @cathay4t can you give some more detail on what you are thinking with your suggestion?

joehandzik:
@tasleson So when you say "separate retrievable value"...are you referring to preferring @cathay4t's suggestion about the by-id path? Or are you saying that you support having both values present, and want to avoid the sort of conversion that I was describing?

tasleson:
@joehandzik Yes, my initial thought is that it's better to just add a new value instead of having the Id having multiple different meanings.  However, I would like to understand @cathay4t  point some more as I could be missing something.

joehandzik:
@tasleson Oh I see what you're saying. I don't think he's advocating that Disk.id have different meanings...I think he's saying he would wholesale replace the current implementation (serial number) with his VPD 0x83 value. For my purposes, as long as the id uniquely identifies the disk it doesn't matter, though that sort of change might be best left to a 2.0 update.

Any thoughts about the code I have in this patch?

tasleson:
@joehandzik OK, I think I'm understanding this all a little more.  You are wanting to put in a value that is /dev/sd*, it's "free" as you say, the utility will dump it out for you.  @cathay4t is saying lets just put the vpd in the id and then potentially do a look-up to get the /dev/sd* path, need to clarify with Gris.  In general /dev/sd* are usually discouraged in use because of their empheral nature.  People get burned by using them all the time, as they change from boot to boot, hot-plug to hot-plug, but you may be using them for a quick look-up/correlation.  Historically the library id is nothing more than a unique string that consistently points to the same resource and they vary from plug-in to plug-in, vendor to vendor, thus they shouldn't really have any other meaning.  If we make the Id the vpd then it can never change going forward and it always must be retrievable for all.  At the moment I'm leaning towards adding the vpd to the disk as an optional attribute (unless we are 100% sure we can retrieve for all, now and in the future) and then provide a vpd to sd path look-up function.  This is because this function could work for both Disks and Volumes, something which would be useful for both.  In some cases the sd look-up will fail to find one because it doesn't exist, eg. Volume which is not masked & mapped, or a disk in a SAN box that isn't physically attached to any system to have a path.  It would also be nice to have functions which take a sd path or vpd and translate back to the Volume or Disk within the library, when applicable too.

joehandzik:
@tasleson yeah I think you've got it (though I thought @cathay4t was trying to nix the dev/sd<> path entirely...we'll let him clarify). I'm with you on avoiding the use of /dev/sd<> for anything more than informational purposes, so I think you're right on the money with what we should shoot for. 

I have 3-4 more patches that I'm going to shoot for posting today as examples of how I'm leveraging certain pieces of data and the other types of functionality that I'm shooting for, and we can rework from there. I have five different waves of these patches, since some patches leverage the data gathered in other patches. I minimized overlap as much as possible.

cathay4t:
@tasleson @joehandzik Example on my thoughts on this:
* LibstorageMgmt provides enough information to help user/client identify the disk,
  like providing VPD83 NAA ID of HBA/JBOD disk.
* A new C and Python standalone library/binding[0], example:
    ```C
//libudev_storage.h

/*
 * Find out the "/dev/sdX" for given VPD83 ID(NAA type prefered).
 */
const char UDEV_sd_path_of_vpd83(const char *vpd83);
    ```
    This is also what I am doing in my SES plugin which require udev information also.

Cons:
* The `libudev_storage.h` might[1] be GPL license as it require full understand of udev(GPLv2+) codes
   and even kernel code(GPLv2).
   If license is the concern, I could code an LGPL or even 4-BSD license library by using SCSI
   standard(like sg3_utils) without checking udev property like `ID_WWN_WITH_EXTENSION`[2].

Pros:
* We don't have to repeat ourselves in every plugins on these methods. As basically, every plugin
   just search sdX using certain ID.

[0]: If preferred  to be in libstoragemgmt project, we could create `libstoragemgmt_scsi.h` and `lsm.SCSI`. Or a standalone library `libudev_storage.h` or `libscsi.h`.

[1]: I am checking with layers about this.

[2]: Actually, by reading the udev code, `ID_WWN_WITH_EXTENSION` is not always VPD83 NAA ID.

tasleson:
Technically we don't need a separate library for translating a VPD to /dev/sd\*, it's after all a library and there is no hard requirement that it's required to talk to a plugin to do meaningful things.  Additionally if we provide functions that take a /dev/sd\* and return a library Volume/Disk they would need to be in the library, so having both in the library seems logical and complementary.

joehandzik:
Ok guys, so...I'd like to know when you get a chance what you'd like me to change about this specific PR, if anything (I'm assuming there's something). It sounds like @tasleson is open to support for what I'm suggesting, but @cathay4t is still a bit wary? Again, my requirements are going to be to be able to avoid udev entirely, so I'm not opposed to separate functionality that touches udev...I just want to be able to avoid it to get the same data.

tasleson:
@joehandzik, @cathay4t The specific use case is how to identify a disk in the library with the block device file like /dev/sda.

In my opinion having a generic way to do this for both volumes & disks is preferred to adding a specific get method to the disk object.  I don't think the implementation should determine what the API looks like.  Implementations can change, the API remains.  This particular raid card offers this for free, but other cards/arrays etc. do not.  To this end we need something to correlate the library disk with the block device.  This is typically done with the vpd 0x83 as you can retrieve this in the data path and out of band with the library.

I'm leaning towards changing this PR to retrieve the disk VPD instead of the sd path.  Then implement a generic translation from VPD to sd path and sd path to VPD or library disk/volume.  I don't think we should try and convert the disk.id to the VPD for reasons I stated earlier.  The implementation of these translation routines could be udev db, walking sysfs, whatever.

Tomorrow @ 8am CDT we have an IRC meeting scheduled on #libstoragemgmt @freenode.net.  Joe if you can attend that would be great to discuss this some more as needed in real-time.

tasleson:
FYI: The license issue that Gris had concerns with is not an issue, verified by legal.

tasleson:
*"Again, my requirements are going to be to be able to avoid udev entirely, so I'm not opposed to separate functionality that touches udev...I just want to be able to avoid it to get the same data."*

@joehandzik No worries with this, Gris or I can crank these translation functions out.

cathay4t:
I create an demo about this:

https://github.com/cathay4t/libstoragemgmt/tree/vpd_2_sd

```
    Demo: API for querying scsi disk name from VPD83 ID.

    C(using libudev):
        int lsm_scsi_disk_name_list_of_vpd83(const char *vpd83,
                                             const char ***sd_name_list,
                                             uint32_t *sd_name_count);

        void lsm_scsi_disk_name_list_free(const char **sd_name_list,
                                          uint32_t sd_count);

    Python(using pyudev):
        lsm.SCSI.disk_name_of_vpd83(vpd83)
```

Due to bug https://bugzilla.redhat.com/show_bug.cgi?id=1273306
if you are using RHEL 7.2, SATA disk will got incorrect result.

joehandzik:
Hey guys, sorry for the late notice but I need to take my son to school this morning @ 8 AM, can't make that meeting slot. Do you guys meet every day at that time, or just on Thursdays? I can probably make tomorrow.

cathay4t:
@joehandzik Very Thursdays. But next one is cancelled.
I will be in the IRC channel tomorrow, hopefully we could all there tomorrow.

cathay4t:
Performance test:
 * RHEL 7.2
 * Has 15k disks[1].
 * Has 1 disk(sda) from hpsa.
 * Try to find out the sda of hpsa disk VPD83 ID.

```
 * C:         0.67s
 * Python:    0.85s
 * hpssacli: 12.42s
```
[1]: 15k disks are simulated by scsi_debug, not from real hardware.

**This indicates udev does not have performance issue on searching VPD83 in large system.
So performance is not a good reason to let plugins to do the VPD83 lookup.**

Raw data:
```
[fge@megadeth libstoragemgmt]$ time -p test/test_vpd83 600508b1001c79ade5178f0626caaa9c
wwn: 0x600508b1001c79ade5178f0626caaa9c
sd_path: /dev/sda
got sda
real 0.67
user 0.22
sys 0.45

[fge@megadeth libstoragemgmt]$ lsmenv sim time -p  test/test_vpd83.py 600508b1001c79ade5178f0626caaa9c
Device alias: sim
URI: sim://
time -p test/test_vpd83.py 600508b1001c79ade5178f0626caaa9c
['sda']
real 0.85
user 0.37
sys 0.48

Time: 0:00.85

[fge@megadeth libstoragemgmt]$ time -p sudo hpssacli ctrl all  show config detail | grep 600508b1001c79ade5178f0626caaa9c -i  -A1
   Primary Boot Volume: logicaldrive 1 (600508B1001C79ADE5178F0626CAAA9C)
   Secondary Boot Volume: logicaldrive 1 (600508B1001C79ADE5178F0626CAAA9C)

--
         Unique Identifier: 600508B1001C79ADE5178F0626CAAA9C
         Disk Name: /dev/sda
real 12.42
user 3.26
sys 9.14
```

cathay4t:
New pull request https://github.com/libstorage/libstoragemgmt/pull/71 providing library level method for this request as we discussed in IRC.

cathay4t:
Feature has been merged to master repo. Closing.

PR #65, opened 2015-10-27
libstoragemgmt: Pull path information for physical disks

joehandzik:

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -346,6 +346,7 @@ class SmartArray(IPlugin):
         cap.set(Capabilities.VOLUME_RAID_CREATE)
         cap.set(Capabilities.SYS_FW_VERSION_GET)
         cap.set(Capabilities.SYS_MODE_GET)
+        cap.set(Capabilities.DISK_LOCATION)
         cap.set(Capabilities.VOLUME_LED)
         return cap
 
@@ -529,13 +530,18 @@ class SmartArray(IPlugin):
         disk_type = _disk_type_of(hp_disk)
         blk_size = int(hp_disk['Native Block Size'])
         blk_count = int(_hp_size_to_lsm(hp_disk['Size']) / blk_size)
+        disk_port, disk_box, disk_bay = disk_num.split(":")
+        disk_location = "Port: %s Box: %s Bay: %s" % (disk_port, disk_box, 
+            disk_bay)
+        
         status = _disk_status_of(hp_disk, flag_free)
         plugin_data = "%s:%s" % (ctrl_num, disk_num)
         vpd83 = LocalDisk.vpd83_get(hp_disk.get('Disk Name', ''))
 
         return Disk(
             disk_id, disk_name, disk_type, blk_size, blk_count,
-            status, sys_id, _plugin_data=plugin_data, _vpd83=vpd83)
+            status, sys_id, _plugin_data=plugin_data, _vpd83=vpd83,
+            _disk_location=disk_location)
 
     @_handle_errors
     def disks(self, search_key=None, search_value=None,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -159,6 +159,8 @@ typedef enum {
     /**^ Plug-in allows user to retrieve storage firmware version */
     LSM_CAP_SYS_MODE_GET = 161,
     /**^ Plug-in allows user to retrieve storage mode*/
+    LSM_CAP_DISK_LOCATION = 163,
+    /**^ Plug-in allows user to retrieve disk location */
     LSM_CAP_VOLUME_LED = 171,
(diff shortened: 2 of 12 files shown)


plugin/hpsa/hpsa.py
```diff
@@ -502,12 +503,26 @@ def _hp_disk_to_lsm_disk(hp_disk, sys_id, ctrl_num, key_name,
         disk_type = _disk_type_of(hp_disk)
         blk_size = int(hp_disk['Native Block Size'])
         blk_count = int(_hp_size_to_lsm(hp_disk['Size']) / blk_size)
+
+        if 'Disk Name' in hp_disk.keys():
+            regex_match = re.compile("/dev/(sd[a-z]+)").search(disk_sd_path)
```
mulkieran: ```disk_sd_path``` doesn't appear anywhere else, I think you must mean something else.

Also, I can't help thinking that the pyudev library would be the right thing to use here.

mulkieran:
Sorry, I see this has been put on the back-burner.

plugin/hpsa/hpsa.py
joehandzik: That's a good catch, it exists in the branch that I cherry-picked from, but does not exist here...I'll figure something out.

Are you suggesting pyudev to read sysfs attributes, or for some other reason? I suppose since it's in the repo now as a dependency, it can comfortably be used. When I created this PR though, it wasn't being pulled in.

Thanks for taking a look!

joehandzik:
@mulkieran It sure seemed that it was back-burner'd, but that wasn't my intention. I honestly just missed the functionality-breaking bug that you pointed out.

joehandzik:
@mulkieran Fixed, I had forgotten that the key I was checking actually gives me the disk path. 

I'm fine with updating libstoragemgmt to use pyudev in situations like this, but there's another chunk of code that does basically the exact same thing: https://github.com/libstorage/libstoragemgmt/blob/master/plugin/hpsa/hpsa.py#L471

We could fix them both at the same time, or fix the existing one first and update this PR to use pyudev too. I'm fine with either option.

cathay4t:
@joehandzik Your patch on hpsa plugin only support JBOD/HBA disks. Could you add support of RAID disks?

joehandzik:
Hey Gris,

Thanks for taking a look. I can see what I can do to support RAID, yeah. I'll shoot to have something done this week.

joehandzik:
@cathay4t Sorry this lagged a bit. I ended up simplifying and eliminating the sysfs attribute usage. You already had the port, bay, and box information via hpssacli. In general, I'd like to move away from using hpssacli, but in this path you're already querying that data so I wasn't buying us anything here. I suppose a move away from hpssacli would be better saved for the ses:// plugin anyway.

joehandzik:
Not sure that the failure here has anything to do with my code...looks like a vpd83 issue. Lemme know if you think it could still be something I did.

tasleson:
@joehandzik I cleaned up the nstor system and it's now passing again.  Not sure why the nstor solution falls down in this case.

joehandzik:
@tasleson Thanks for taking a look! Glad I'm not crazy. :)

joehandzik:
@cathay4t Lemme know if this one is ok too!

joehandzik:
The nstor plugin clearly hates me.

tasleson:
@joehandzik The test VM in the lab is having some weird issues, slow downs hangs etc.  I'm in the middle of trying to sort things out.  I'm pretty sure no changes you have made here are breaking nstor.

joehandzik:
@cathay4t I rebased this one a few days ago to get it merge-able. Any further changes required, or can it be merged?

tasleson:
@joehandzik Couple of questions:
1. What exactly do you get in the location string?
2. Is this functionality supported for any other hardware?

joehandzik:
@tasleson 

1. Port, Bay, and Box values for the disk. So, a user can pull each physical disk's exact location. This could be leveraged for a GUI or CLI interface to communicate the exact location of a disk in a chassis (at scale, that's valuable information). It's actually already embedded in an existing variable, but it's not spelled out exactly which piece of the string corresponds to port, bay, or box. Now there's no guesswork on the part of the user.

2. I've been burned once by saying this, but I feel more confident this time: At the very least, MegaCLI/Perccli should expose this data. I have limited experience with SANs (and zero real experience managing them), so I don't know if that's data an admin would want in EMC/Netapp/3PAR environments.

tasleson:
@joehandzik Thanks for the information.  Yes, I agree knowing where the disk is very helpful.

The only question I have is if this change supports RAID and not just JBOD per Gris' inquiry on 3/11?  I'm unable to determine this by looking at the change.  I'm assuming that when used in a RAID configuration the user would select the volume of interest, use the pool information to query which disks comprise the pool `pool_member_info` and then use this API call to determine the location for each disk?

joehandzik:
@tasleson Your assumption is correct, and the change I made removed the conditional that only populated this path value when the controller was in JBOD mode. So it'll safely work in both modes. :)

tasleson:
Thanks @joehandzik !

PR #69, opened 2015-10-30
Add volume-ident-led-* support for volumes

joehandzik:
Hopefully this one is less controversial. These two patches allow you to enable and disable LEDs behind logical volumes on a Smart Array (and I assume other controllers, but I don't know the other plugins well enough to implement for them). Take a look lemme know what you think.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt.h
@@ -1017,6 +1017,18 @@ int LSM_DLL_EXPORT lsm_volume_ident_led_set(lsm_connect *c,
                                             lsm_volume *volume,
                                             lsm_flag flags);
 
+/**
+ * Disable the IDENT LED for the desired volume.
+ * New in version 1.3, only available for hardware RAID cards.
+ * @param[in] c				Valid connection
+ * @param[in] volume			A single lsm_volume
+ * @param[in] flags			Reserved, set to 0
+ * @return LSM_ERR_OK on success else error reason.
+ */
+int LSM_DLL_EXPORT lsm_volume_ident_led_clear(lsm_connect *c,
+                                              lsm_volume *volume,
+                                              lsm_flag flags);
+
 #ifdef  __cplusplus
 }
 #endif
--- a/c_binding/include/libstoragemgmt/libstoragemgmt.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt.h
@@ -1005,6 +1005,18 @@ int LSM_DLL_EXPORT lsm_volume_raid_create(lsm_connect *c,
                                           lsm_volume **new_volume,
                                           lsm_flag flags);
 
+/**
+ * Enable the IDENT LED for the desired volume.
+ * New in version 1.3, only available for hardware RAID cards.
+ * @param[in] c				Valid connection
+ * @param[in] volume			A single lsm_volume
+ * @param[in] flags			Reserved, set to 0
+ * @return LSM_ERR_OK on success else error reason.
+ */
+int LSM_DLL_EXPORT lsm_volume_ident_led_set(lsm_connect *c,
+                                            lsm_volume *volume,
+                                            lsm_flag flags);
+
 #ifdef  __cplusplus
 }
 #endif
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -159,6 +159,8 @@ typedef enum {
     /**^ Plug-in allows user to retrieve storage firmware version */
     LSM_CAP_SYS_MODE_GET = 161,
     /**^ Plug-in allows user to retrieve storage mode*/
+    LSM_CAP_VOLUME_LED = 171,
+    /**^ Plugin allows user to set and clear volume LEDs */
     LSM_CAP_POOLS_QUICK_SEARCH = 210,
     /**^ Seach occurs on array */
     LSM_CAP_VOLUMES_QUICK_SEARCH = 211,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -981,6 +981,18 @@ typedef int (*lsm_plug_volume_raid_create) (lsm_plugin_ptr c,
                                             lsm_volume ** new_volume,
                                             lsm_flag flags);
 
+/**
+ * Enable the IDENT LED for the desired volume.
+ * New in version 1.3, only available for hardware RAID cards.
+ * @param[in] c			Valid lsm plug-in pointer
+ * @param[in] volume		A single lsm_volume
+ * @param[in] flags         	Reserved, set to 0
+ * @return LSM_ERR_OK on success else error reason.
+ */
+typedef int (*lsm_plug_volume_ident_led_set) (lsm_plugin_ptr c,
+                                              lsm_volume * volume,
+                                              lsm_flag flags);
+
 /** \struct lsm_ops_v1_2
  * \brief Functions added in version 1.2
  */
@@ -990,6 +1002,15 @@ struct lsm_ops_v1_2 {
     lsm_plug_pool_member_info pool_member_info;
     lsm_plug_volume_raid_create_cap_get vol_create_raid_cap_get;
     lsm_plug_volume_raid_create vol_create_raid;
+};
+
+/** \struct lsm_ops_v1_3
+ * \brief Functions added in version 1.3
+ * NOTE: This structure will change during the developement util version 1.3
+ *       released.
+ */
+struct lsm_ops_v1_3 {
+    lsm_plug_volume_ident_led_set vol_ident_set;
 };
 
 /**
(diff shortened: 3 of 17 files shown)


cathay4t:
Two ideas:
1. Can we add a property to disk indicate current LED status?
2. How about use on/off instead of set/clear? I am not native speaker, not sure which one is better.

joehandzik:
1. I think that should be a separate API. I am pretty sure hpssacli cannot (or does not, anyway) give a user that data. I'll double check that, but all the same, I'd expect an LED status call to be a separate API implementation. For hba mode disks, you can get that state information via a page inquiry.

2. I went with the same terminology used in sg_ses. I consider it to be a far more standard implementation than any other tool. See this:

$ sg_ses --dev-slot-num=0 --set=ident /dev/sg3
$ sleep 10
$ sg_ses --dsn=0 --clear=ident /dev/sg3

For more information: http://sg.danny.cz/sg/sg_ses.html

Make sense? For physical disks, we can implement at the library level via sg_ses, similar to what we discussed for device paths in today's IRC chat. Sound like a plan?

cathay4t:
Yeah. My intention is:
* SES plugin handle all hpsa/megaraid HBA/JBOD mode and other HBA/JBOD enclosures(mtp2sas/mpt3sas). Stay tune, coming soon.
* hpsa plugin and megraid plugin handle RAID mode.

cathay4t:
For library level code about physical disk, I need some time to think it through.
I don't know how to implement it and why we need it at library level yet.

joehandzik:
Hey @cathay4t, any updated thoughts here? If not, I'd really like to get this pulled in so I can rebase my fork. I assume, at the very least, I need to rebase this branch to pull in the latest changes in master. Anything else?

joehandzik:
Just completed a rebase with the current master, so...hoping that this one can get pulled in.

cathay4t:
I am wondering why on volume level instead of disk level?
   * `lsm_disk_ident_led_set()`
   * `lsm.Client.disk_ident_led_set()`

If user want to set LEDs on all physical disk of certain volume, they could use `lsm.Client.pool_member_info()` to find out disk ownership and set LED of these 
physical disks.

Please give me another week to find out the SES standard, if possible, 
I would like introduce LED status querying method before this one.

joehandzik:
Not possible behind a Smart Array in RAID mode, not to mention obnoxious. Smart Array, when in RAID mode, can enable LEDs on both an entire logical volume as well as individual physical drives. But, when we have the ability to enable the entire volume with a single command, why wouldn't we do it? Also, like I've said before, when in RAID mode you won't be able to query for the page that would tell you if the drive's LED is enabled.

I feel like I've seen you request both more and less complexity in all of the PRs I've proposed. I'm starting to get a bit frustrated because it seems like you request the opposite of what I've proposed every time (for example, what you're suggesting here is needlessly complicated in my opinion). Maybe we need to take a step back and make sure we're on the same page about the goals here? The more I've looked at your implementation in #71, the more I think it'll add complexity to the host application (my PR set in my wip-all branch keeps every piece of metadata necessary to blink and LED on a physical drive, for example...I track the SEP with the disk to enable simple commands from the user level to blink the LED).

Would appreciate some thoughts from @tasleson too, if he has the cycles. Appreciate all the interaction so far!!

tasleson:
My initial reaction was kind of like @cathay4t, *"What's with blinking all the lights for a volume?"*.  My use case is finding a disk that needs to be replaced.  However, I don't see any reason why adding blinking for all the disks that make up a volume is bad from an API decision.  It should work fine for a HBA or a SAN box that has 1K+ disks, in fact in larger configurations it becomes more of an asset if needed (This virtual disk sits on 1K disks, light them up so we can remove them for law enforcement).  My main thought when looking at the API change is if the change works well across all the different things we manage?  In this case I think having it on the Volume is perfectly acceptable too.

Additionally, we need to remember that it could be possible that an array could implement blinking lights for a volume, but not support `pool_member_info`, so in some cases a user couldn't implement this functionality themselves.

joehandzik:
That's fair. My use case is at scale to locate the volume in the first place. I'm looking to implement RESTful interfaces into this library through various solutions that I work with, so my use case is: "My Ceph OSD is failing, and I want to locate the logical volume I'm dealing with to service it amidst several racks of servers". Also, generally, the volumes will be single disk RAID 0 volumes for scale-out use cases (or better yet, physical disks behind HBA mode controllers), so the LEDs being enabled would be the physical service unit anyway.

It's fair to say though that a physical disk might be the service level, which to me makes it essential to implement the LED enabling functionality for physical disks in the plugins themselves. For Smart Array, for example, the right way to do this is probably:

1. Check to see if the controller is in HBA mode or RAID mode
2a. If RAID mode, use hpssacli to enable the disk's LED
2b: If HBA mode, use sg_ses or some other direct interface into the SEP to enable the disk's LED

I really don't have an informed opinion about if what I'm proposing works for all supported hardware...if I need to have that to contribute, I'll look over the code. But so far, I'm been relying on you and @cathay4t to keep me honest with respect to other supported hardware.

Thanks, @tasleson!

cathay4t:
@joehandzik Sorry I was quite impolite on your PRs.

Ideally, yes, the use case sounds reasonable and we could have it.

By looking on other plugins, it seems hpssacli is the only tool supporting volume LED function:

 * SMI-S plugin

DMTF has DSP1074 "Indicator LED Profile" allows flash LED on volume[1].
EMC is the only vendor I found supporting DSP1074 and only support disk/slot LED.

* Avago/LSI MegaRAID
Only support disk level LED function vis storcli.

* NetApp plugin.
No SDK API found for volume or disk LED. Only found CLI command `storage disk set-led`.

* Targetd plugin
No support.

* Nstor plugin
Only support JBOD disk indicator.

* SES plugin
Only support disk LED.

So the questions come to:

Do we introduce this API interface which only has 1 plugin implemented while other
plugin raise no support error or implemented via `pool_member_info() -> disk_ident_led_set()` way?



[1]: The standard defines LED could associated to CIM_ManagedSystemElement where 
      CIM_StorageVolume (volume) are CIM_PhysicalPackage(disk) both inherit from.

joehandzik:
@cathay4t I didn't think you were being particularly rude about anything, I just wanted to make sure you understood where I was coming from. Thanks for clarifying the concerns you have!

I actually don't have much familiarity with SMI-S, PMC doesn't implement it. But given that some vendors have implemented a mechanism to flash LEDs behind an entire volume, I'd like to allow for using that functionality. Frankly, it'll help give me some runway to not need to operate with a forked libstoragemgmt in the Ceph project...as long as I can blink LEDs behind logical volumes, I have enough to put together functionality demos. :)

tasleson:
@cathay4t  Thanks for digging into the different plug-ins & arrays to see what's possible.  We have other functionality which is only supported by one array at the moment, so adding this for one HBA doesn't seem like much of a stretch.  We could add this support to targetd so then we could have two that support it.  If a plugin supports `pool_member_info` I say we just implement it on the plugin side with what you were proposing end users could do.  If @joehandzik is going to wrap all of this up with some RESTful API he will appreciate not having to do the round trips to flash all disks that make up a Volume.   Adding an API call that many plug-ins do not support doesn't require touching all the plugins (well it's not suppose too, and I hope that's still a correct statement, otherwise we broke it).

Like I stated before, the API add seems reasonable and logical.

joehandzik:
@tasleson It's still accurate that adding an API call that is only supported by a single plugin only requires adding code for that single plugin. You'd just never advertise support for the API call within the plugin itself.

I assume the pun was fully intended (re: "API...logical"). ;P

cathay4t:
@joehandzik The plugin implements the `lsm.Client.capabilities()`(python) or `lsm_capabilities()`(C) to indicate so, we will make sure existing plugins does not return inaccurate data.

Then we all agree on introducing this new API.
While I am reviewing this PR and working on implementation of MegaRAID,
do you(@joehandzik) have disk LED set/clear PR in hand?
If not, I could coded them out.

cathay4t:
If possible, please include manpage for lsmcli about new command `volume-ident-led-set` and `volume-ident-led-clear`.

joehandzik:
Ok, I addressed the bits in hpsa.py and _client.py that you pointed out. Not sure I understand why we'd converge the separate unit tests...some clarity there would be appreciated.

joehandzik:
Also, regarding lsmcli manpage...should I just edit the lsmcli manpage in doc/man directly? Or is there a better mechanism for that?

cathay4t:
For manpage, edit `/doc/man/lsmcli.1.in` should be enough.

For unit test, you could just create one test case run set and clear since they are came from the same capability `LSM_CAP_VOLUME_LED`. Anyway this is trivial changes, you can still keep using
two test cases if you want.

joehandzik:
I prefer to leave them separate since I added the functionality with two separate commits. If necessary, I can add two LSM_CAP_VOLUME_LED support enums instead of one, but I think it makes sense the way I wrote it. So if you don’t feel strongly, I’ll leave it as-is.

I will update the man page documentation today, hopefully that will be enough to start pulling things upstream.

Joe

joehandzik:
Alright, documentation is a go!

joehandzik:
@tasleson @cathay4t Is this PR good to go?

cathay4t:
@joehandzik Sorry for the late response. Code review looks good, let me run some test before commit. Thanks.

joehandzik:
It's all good, lemme know how your testing goes. I have validated on a Gen9 DL380 using an H240ar in RAID mode.

cathay4t:
The patch does not apply clean with current mater:
```
error: patch failed: test/plugin_test.py:1385
error: test/plugin_test.py: patch does not apply
Patch failed at 0001 libstoragemgmt: Add volume-ident-led-set to enable LEDs on all disks in a logical volume
```
The patch command indicates:
```
patching file test/plugin_test.py
Hunk #1 succeeded at 1423 with fuzz 1 (offset 38 lines).
```

It seems github can handle this automatically. But I would suggest a rebase to current master.

cathay4t:
The `NOT_FOUND_VOLUME` error still not handle correctly in simulator plugin and hpsa plugin:
 * simulator plugin return success on not exist volume.
 * hpsa plugin return `PLUGIN_BUG`.

To test this, you could refer to `test_volume_raid_info()` of `plugin_test.py`.

Let me know if you need help on implement this.

joehandzik:
Hey Gris,

I fixed these items:

1. Removed the unexpected newline bits that you mentioned (for some reason I had injected newlines between functions in the wrong patch)
2. Check for volume existence in sim plugin
3. In hpsa plugin, it's a multifaceted approach. Do not change execution of LED enable/disable at all, but in the event of a failure actually search for the volume in hpssacli output. If you find the volume, it's a PLUGIN_BUG. If you don't find the volume, it's a missing volume instead.
4. Change CLI implementation to the method that you requested.

Hope that covers it!

joehandzik:
Thanks for catching that, can't believe I didn't update the error messaging on those two.

cathay4t:
Looks good. Merged.

PR #71, opened 2015-11-03
[PATCH V9] New function for SCSI disk path matching.

cathay4t:
* API for querying scsi disk paths from VPD83 ID.
```
    C:
        int lsm_scsi_disk_paths_of_vpd83(const char *vpd83, 
                                         lsm_string_list **sd_path_list, 
                                         lsm_error **lsm_err)

    Python:
        lsm.SCSI.disk_paths_of_vpd83(vpd83)
```

* lsmcli changes: add "Disk Paths", example:
```
ID           | Name  | SCSI VPD 0x83    | Size       | Disabled | Pool ID       | System ID | Disk Paths
--------------------------------------------------------------------------------------------------------
VOL_ID_00001 | test  | 53333330000007d0 | 1073741824 | No       | POOL_ID_00001 | sim-01    | /dev/sda
             |       |                  |            |          |               |           | /dev/sdf
             |       |                  |            |          |               |           | /dev/sdh
             |       |                  |            |          |               |           | /dev/sdj
VOL_ID_00002 | test2 | 53333330000007d1 | 1073741824 | No       | POOL_ID_00001 | sim-01    | /dev/sde
             |       |                  |            |          |               |           | /dev/sdg
             |       |                  |            |          |               |           | /dev/sdi
             |       |                  |            |          |               |           | /dev/sdk
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ addons:
         - python-m2crypto 
         - libssl-dev 
         - libconfig-dev
+        - libudev-dev
+        - python-dev
 
 compiler: gcc
 
--- a/c_binding/Makefile.am
+++ b/c_binding/Makefile.am
@@ -3,13 +3,15 @@ SUBDIRS = include
 AM_CPPFLAGS = 	-I$(top_srcdir)/c_binding/include \
 				-I$(top_builddir)/c_binding/include \
 				-I@srcdir@/c_binding/include \
-				$(LIBXML_CFLAGS) $(LIBGLIB_CFLAGS)
+				$(LIBXML_CFLAGS) $(LIBGLIB_CFLAGS) \
+				$(LIBUDEV_CFLAGS)
 
 lib_LTLIBRARIES = libstoragemgmt.la
 
-libstoragemgmt_la_LIBADD=$(LIBXML_LIBS) $(YAJL_LIBS) $(LIBGLIB_LIBS)
+libstoragemgmt_la_LIBADD=$(LIBXML_LIBS) $(YAJL_LIBS) $(LIBGLIB_LIBS) \
+			 $(LIBUDEV_LIBS)
 libstoragemgmt_la_LDFLAGS= -version-info $(LIBSM_LIBTOOL_VERSION)
 libstoragemgmt_la_SOURCES= \
 	lsm_mgmt.cpp lsm_datatypes.hpp lsm_datatypes.cpp lsm_convert.hpp \
 	lsm_convert.cpp lsm_ipc.hpp lsm_ipc.cpp lsm_plugin_ipc.hpp \
-	lsm_plugin_ipc.cpp util/qparams.c util/qparams.h
+	lsm_plugin_ipc.cpp util/qparams.c util/qparams.h lsm_scsi.c
--- a/c_binding/include/libstoragemgmt/Makefile.am
+++ b/c_binding/include/libstoragemgmt/Makefile.am
@@ -23,6 +23,7 @@ lsminc_HEADERS =			\
    libstoragemgmt_targetport.h          \
    libstoragemgmt_types.h		\
    libstoragemgmt_version.h             \
+   libstoragemgmt_scsi.h		\
    libstoragemgmt_volumes.h
 
 
--- a/c_binding/include/libstoragemgmt/libstoragemgmt.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt.h
@@ -34,6 +34,7 @@
 #include "libstoragemgmt_systems.h"
 #include "libstoragemgmt_targetport.h"
 #include "libstoragemgmt_volumes.h"
+#include "libstoragemgmt_scsi.h"
 
 
 /*! \mainpage libStorageMgmt
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_scsi.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_scsi.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2015 Red Hat, Inc.
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Gris Ge <fge@redhat.com>
+ *
+ */
+
+#ifndef LIBSTORAGEMGMT_SCSI_H
+#define LIBSTORAGEMGMT_SCSI_H
+
+#include "libstoragemgmt_common.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Find out the scsi disk paths of given SCSI VPD page 0x83 NAA ID.
+ * New in version 1.3.
+ * @param[in] vpd83 String. The VPD83 ID retrieved from LSM volume or disk.
+ *                  for supported strip sizes.
+ * @param[out] sd_path_list
+ *                  Output pointer of lsm_string_list. The format of scsi
+ *                  disk path will be "/dev/sd[a-z]+".
+ *                  NULL if no found or got error.
+ *                  Memory should be freed by lsm_string_list_free().
+ * @param[out] lsm_err
+ *                  Output pointer of lsm_error. Error message could be
+ *                  retrieved via lsm_error_message_get(). Memory should be
+ *                  freed by lsm_error_free().
+ * @return LSM_ERR_OK                   on success.
+ *         LSM_ERR_INVALID_ARGUMENT     when any argument is NULL.
+ *         LSM_ERR_NO_MEMORY            when no memory.
+ *         LSM_ERR_LIB_BUG              when something unexpected happens.
+ */
+int LSM_DLL_EXPORT lsm_scsi_disk_paths_of_vpd83(const char *vpd83,
+                                                lsm_string_list **sd_path_list,
+                                                lsm_error **lsm_err);
+
+#ifdef __cplusplus
+}
+#endif
+#endif                          /* LIBSTORAGEMGMT_SCSI_H */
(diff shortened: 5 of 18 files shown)


tasleson:
@cathay4t Please update the travis file with the new dependencies.

tasleson:
Is there some reason you are removing the /dev?  I'm thinking the client and their use case will need to pre-pend /dev/ to actually use the value?

joehandzik:
Seconded, @tasleson.

tasleson:
Would these names be better?
`lsm_vpd_to_scsi_disk_names`
`lsm.SCSI.vpd_to_scsi_disk_names`

cathay4t:
For NULL vs empty `lsm_string_list`, all existing C functions use NULL:
 * `lsm_available_plugins_list()`
 * `lsm_pool_member_info()`
 * `lsm_nfs_auth_types()` [1]

For None vs empty list, all existing python functions use empty list as the `@_return_requires()` enforced.

[1]: It seems we don't have implementation of `lsm_nfs_auth_types()`, unless I missed anything.

cathay4t:
The API name about `scsi_disk_name`, not `scsi_disk_dev_path`, so I removed the `/dev/` prefix.

cathay4t:
Travis CI configure file updated, but waiting robot to add `python-pyudev` pacakge into whitelist:
https://github.com/travis-ci/apt-package-whitelist/issues/1656

cathay4t:
@tasleson The function name you suggested cause python one has two `scsi`, and maybe vpd83 is better fit due to existence of vpd page 0x80.

joehandzik:
@cathay4t If that's the way you're thinking about it, then I'd prefer to rename the API call to scsi_disk_dev_path and include the path. The node by itself is not really required, and I'd rather give users the full path and let them chop it down than give them a partial path. Another good reason for that is for block devices that would exist in dev/block...better to give the whole path for context, IMO.

tasleson:
I don't want to get bogged down in the minutiae, we don't need to get caught up `"bikeshedding"` here :-).  I think we are mostly in agreement here.  My preference would be to return the full path to the device.  For an end user of the API I believe that's what they are going to use it for, so why make them jump though additional hoops to use it.  If that requires a different name to clarify, so be it.  I do think we should be specific and state vpd 0x83 as there are other vpd pages.  So unless we can come up with a good reason why returning the full path is bad, lets do the full path.

With this change we need to return the vpd for each of the disks.  I don't think I've seen a PR on that?

joehandzik:
Thanks @tasleson! Had to look up bikeshedding. ;P 

Definitely agree on this one. I think my concern here is the workflow. In my implementation, the /dev/sd<> path is an element of the disk for python, which it isn't in this case. Definitely a design choice, but it looks like the future workflow for, say, blinking a physical disk LED becomes more complex than what I would desire:

@cathay4t's implementation (extrapolating based on this PR):
-lookup SAS address or bay index for a physical disk via call into SCSI library
-lookup physical disk's SEP via call into SCSI library
-enable disk's LED via call into SCSI library (with the user needing to bring all the above arguments along for the ride)

My implementation (https://github.com/HP-Scale-out-Storage/libstoragemgmt/commit/42e0be5c93920c585ef278d12afaeb961a994ab7#diff-547822719de91a169b85b2c4b56c837aR765):
-enable disk's IDENT LED via single _client.py command (all required data is contained within the python class)

I understand the desire to minimize repeated code (theoretically, a similar/identical implementation would work for an Avago controller), but maybe there's a better place to put the potentially repeated code that would exist in these cases?

cathay4t:
The use of SAS address in my SES plugin does not mean blinking LED need to go through that.
Because SES-3 SPEC does not provide VPD83, I need SAS address or FC WWN or NVMe serial
to match the udev/kernel disks.

For sure, the LED related functions will be like:
`lsm.Client.disk_indent_led_set(lsm.disk)`.

I believe we are on the same page about this vpd83 to scsi disk path PR. 
I will submit the V2 patch set soon.

cathay4t:
@joehandzik Sorry, I didn't fully understand your points on SES plugin implementation before.

The reason I introduce new plugin SES is to support HBA/JBOD cards like mpt2sas/mpt3sas 
and other T10 SES-2+ standard following cards in a open source way.

Introducing the SES plugin does not `minimize repeated code` or replace hpsa plugin in HBA mode, but expand the hardware support range of libstoragemgmt. Both hpsa and megaraid plugin should
implement disk and/or volume LED functions using their binary tools in order to support RAID mode.

It's user's choice to use hpsa plugin or SES plugin when come to Smart Array HBA mode.

cathay4t:
V2 patch set uploaded.

cathay4t:
Changes in V2:

* API name changed to `lsm_scsi_disk_paths_of_vpd83()` and `lsm.SCSI.disk_paths_of_vpd83()`.
* Fixed dangling pointer of `sd_path_list`.
* Add test for dangling pointer of `sd_path_list`.
* Changed `lsmcli` output name as `Disk Paths`.

cathay4t:
I am working on the VPD83 property of lsm.Disk in another branch.
Will submit to review after this PR.

cathay4t:
Please hold this PR.
I managed to extra VPD83 NAA ID from `/sys/block/sda/device/vpd_pg83`, if the performance of that way looks good, we could have a udev-free implementation.

joehandzik:
Hey Gris,

Please be aware that not all drivers/controllers will support this sysfs attribute. For example, most Smart Array combos don’t, so this isn’t a safe assumption for HPE’s purposes.

In general though, if you find a sysfs attribute that is reliably present across all supported devices, I’m fine with using it.

Joe

cathay4t:
Both the hpsa and megaraid_sas driver provide this sysfs entry in RHEL 7.2.
In fact, base on my poor understanding of linux kernel code, the sysfs vpd_pg83 is provided by scsi layer, as long as device driver allows VPD83 query[1] IOCTL, that sysfs entry should be valid.

But a downside is, RHEL 6.7 kernel does not support this file. Let me test the performance today
to say how much we get and how much we lose.

[1]: The udev also use VPD83 IOCTL to assemble ID_WWN_WITH_EXTENTION.

cathay4t:
OK. The performance came out as very promising(even better than udev): [1]
```
[fge@megadeth ~]$ ls -l /sys/block/|grep sd |wc -l
15950
[fge@megadeth ~]$ time ./abc  600508b1001c79ade5178f0626caaa9c
Gotcha sda

real    0m0.214s
user    0m0.029s
sys     0m0.185s
```

So, I will use the kernel sysfs way first, if nothing found, use udev as fallback.
The kernel sysfs way only check VPD83 NAA ID which is exactly what we are looking for.
The udev ID_WWN_WITH_EXTENSION may contain other type of VPD83 ID(TGT type has hight priority than NAA).

Stay tune for V3 path set, I will also include `lsm.Disk.vpd83` in this as Tony mentioned.

[1]: Just a simple C tool using kernel sysfs to search sd on vpd83, not in lsm library yet. So the real
performance might be slightly worse.

cathay4t:
New patch set uploaded, but not fully tested. Please don't commit yet.

cathay4t:
V3 patch uploaded.

Changes:
 * Add `lsm_error ** lsm_err` argument to received error message.
 * Try sysfs vpd83 file fist, if not support, use udev.
 * Change python module to C extension of C API.

Since this patch set has 10 patches, I would like to create new PR about `lsm.Disk.vpd83` after this one approved.

cathay4t:
Hold on. RHEL 6 does not ship pkconfig pc file in python-devel. Let me find a workaround for this.

cathay4t:
V4 patch set uploaded.

Changes in V4:
* Fix autotool on RHEL 6 which does not ship pkgconfig file for python-devel.

joehandzik:
Hey @cathay4t, I've got some time this week (libstoragemgmt integration into Ceph ended up not being difficult at all), is there anything you could use help on to move this along? Being able to identify lsm_disk or lsm_volume by its dev path is going to be key once I'm ready to submit a PR into Ceph.

cathay4t:
@joehandzik This is my top priority work this week. 
Besides this PR, that remains adding `lsm.Disk.vpd83`, it request extra care as hpsa does not expose so via hpssacli. I will create library level function `lsm_scsi_vpd83_of_sd_path()` to plugin to unitilize.
So I would use some help on code review and test once this PR rebased and new PR for `lsm.DIsk.vpd83` created.

cathay4t:
Based to current master.

cathay4t:
I would like to keep using `PyList` as [python website][1] mentioned:
```
Tuples are immutable, and usually contain an heterogeneous sequence of elements.
...
Lists are mutable, and their elements are usually homogeneous and are accessed by 
iterating over the list.
```

The scsi disk paths here are homogeneous and should be accessed by iteration.

[1]: https://docs.python.org/2/tutorial/datastructures.html#tuples-and-sequences

cathay4t:
Please hold this new uploaded patch set. Not finish yet still have problem when `make rpm`, will fix it tomorrow.

cathay4t:
V5 patch set uploaed.
Changes in V5:
* Split lsmenv changed to a standalone patch and use Tony suggested way to discovery existing python C 
  extension and link it out.
* Fix spell errors.
* Create new rpm sub package `libstoragemgmt-python-clibs` to hold the `lsm.SCSI` module.
* Fix possible memory leak in python C extension `lsm.SCSI`.
* Use stack instead of malloc in C library function.

cathay4t:
Please wait for the V6 patch set. I missed to check input VPD83 length.

cathay4t:
V6 patch set uploaded.

Changes in V6:
1. Check input VPD83, should not exceeded the max length of vald VPD83 NAA ID(32 + 1).
2. Set `lsm_error` if possible.

tasleson:
@cathay4t  Before I got too involved with the review today I compiled and tried it out interactively in python and wasn't getting any results which lead me to the bug I found and commented on.  Please fix and do a quick sanity test to ensure this is working for both sysfs and udev code paths.  I will then give the code a good review.

Thanks!

cathay4t:
@tasleson Sorry I forgot the retest on RHEL6 after I add vpd83 string check.
Fixed in V7 patch set.

cathay4t:
V7 patch set uploaded.

Changes in V7:
* Fix query vpd83 on RHEL6.
* Fix lsmenv on RHEL 6.

cathay4t:
V7 patch still have issue:

`_udev_vpd83_of_sd_name()` and  `_udev_vpd83_of_sd_name()` should return `LSM_ERR_NOT_FOUND_DISK` when disk not found, instead of `NO_SUPPORT`.

Since all disk string are from `_sysfs_get_all_sd_names()`, user will not trigger this.

Will change it in V8 after Tony reviewed this patch set.

tasleson:
@cathay4t I prefer to re-review after you have known issues resolved, thanks!

cathay4t:
V8 patch set uploaded.

Changes in V8:
* Fix typo.
* Fix `LSM_ERR_NOT_FOUND_DISK` handling.
* Add new test for non-exist vpd83 check.

cathay4t:
V9 patch set uploaded.

Changes in V9:
* Replace library static variable `_lsm_err_msg` with function local variable `char err_msg`.

PR #72, opened 2015-11-04
Demo: New SES Plugin: SCSI enclosure service.

cathay4t:
Demo yet. Please don't commit/merge.

Done:
    * System and disk querying.

TODO:
    * Handle FC and NVMe disks(at least raise error for FC).
    * Polish the code to generate better error message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
         - python-m2crypto 
         - libssl-dev 
         - libconfig-dev
+        - libsgutils2-dev
 
 compiler: gcc
 
--- a/config/Makefile.am
+++ b/config/Makefile.am
@@ -1,11 +1,11 @@
 lsmconfdir=$(sysconfdir)/lsm
 lsmconf_DATA=lsmd.conf
 
-EXTRA_DIST= lsmd.conf pluginconf.d/sim.conf
+EXTRA_DIST= lsmd.conf pluginconf.d/sim.conf pluginconf.d/ses.conf
 
 pluginconfdir=$(sysconfdir)/lsm/pluginconf.d
 
-pluginconf_DATA=pluginconf.d/sim.conf
+pluginconf_DATA=pluginconf.d/sim.conf pluginconf.d/ses.conf
 
 if WITH_MEGARAID
 pluginconf_DATA += pluginconf.d/megaraid.conf
--- a/config/pluginconf.d/ses.conf
+++ b/config/pluginconf.d/ses.conf
@@ -1,0 +1,1 @@
+require-root-privilege = true;
--- a/configure.ac
+++ b/configure.ac
@@ -259,6 +259,14 @@ if test "x$with_systemdsystemunitdir" != xno; then
 fi
 AM_CONDITIONAL(HAVE_SYSTEMD, [test -n "$with_systemdsystemunitdir" -a "x$with_systemdsystemunitdir" != xno ])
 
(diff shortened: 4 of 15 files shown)


cathay4t:
@joehandzik The SES plugin is almost ready for review. 
Could you give it a try on your SmartArray HBA mode?
Thank you.

tasleson:
@cathay4t how exactly does a user leverage the SES plug-in?  Am I correct to conclude that a user would be required to use both this plugin and another to completely manage their HBA based storage (provision and monitor the same hardware)?

cathay4t:
@tasleson The SES plugin mainly is for query, actually it seems no management required for
SCSI enclosure besides flash LED and change disk/slot alias since OS has direct access to these
physical disks.

The future feature of SES plugin might help clarifying this:
 * SES plugin maintain lsm.Disk.status to reflect both SES link status and SMART health. [1]
 * User could flash the identification or fault LED on certain lsm.Disk.
 * Expose more information about lsm.Disk like:
  * Addresses. (SAS address or FC WWN or PCI-E slot name).
  * RPM (0 for SSD, -1 for unknown)
  * Link speed.
  * Port protocol(SAS/SCSI/FC/NVMe).

[1] We might require a new function to query root cause of STATUS_ERROR, TBD.

cathay4t:
And also, SES provides battery, fan, sensor, power supply informations we might utilize in the future.

tasleson:
Is there any reason we couldn't leverage SES within the exiting plugin(s) to provide this functionality without requiring the use of two separate plugins to fully manage the same storage?

cathay4t:
SCSI enclosure like mpt2sas/mp3sas does not fit into hpsa plugin or megaraid plugin, but they do support SES quit well.

Regarding hpsa and megaraid, when configured as HBA/JBOD mode by admin, user are suggested to
use SES plugin.

joehandzik:
Shouldn't mpt2sas work with the megaraid plugin? I suppose I don't understand exactly how MegaRAID works, but for HPE Smart Array products you can still manage the storage via hpssacli for some operations (hence my Frankenstein approach to getting a full picture).

I personally would prefer what @tasleson is suggesting. Why not make the SES functionality part of a library of calls that either the HPE or LSI/Avago plugins could leverage? I'm all for your very optimistic assumption that all storage controller that claim support for SES fully support it, but I can promise you that there are Smart Array firmware versions released today (and expander firmware, for that matter) that imperfectly support SES. The imperfections are certainly different than they would be for other controller vendors. That's another good reason to allow me to add functionality into the HPE plugin within libstoragemgmt...we'll get a better support matrix if each vendor can account for behavior differences. 

Now, I do still think that the SES plugin makes sense as an aspirational goal, but if we pull SES functionality out into a library each plugin can pick and choose what will work and what won't work. I'm pushing PMC to get all the way to an SES-complaint approach...just not there in the field yet. Make sense?

cathay4t:
The mpt2sas or mpt3sas does not works under MegaRAID binrary tool storcli, and still want to keep this open source plugin for those SES-complaint cards.

For sharing SES functions with hpsa and megaraid, I will investigate the possibility of exposing
library level(both C and Python) functions to both user and plugin like:
 * `lsm_ses_ident_led_set("/dev/sda")
 * `lsm_ses_ident_led_clear("/dev/sda")
 * `lsm_ses_fan_list_get(lsm_fan ***fan_list, uint32_t *fan_count)
 * and etc.

How's that sounds?

cathay4t:
Closing this PR as functions here will be changed to library level API.
Any further discuss, please use issue #91

PR #87, opened 2016-01-26
Master udev

mulkieran:
Suggestion for using libudev a bit more in lsm_scsi.c.

1. The first patch is pretty straightforward.
2. The second patch only makes sense if there is no objection to using libudev to look up a sysfs attribute.
3. The third patch only makes sense if the second patch makes sense, so I did it only partially.

### DIFF
--- a/c_binding/lsm_scsi.c
+++ b/c_binding/lsm_scsi.c
@@ -350,11 +350,8 @@ static int _udev_vpd83_of_sd_name(char *err_msg, const char *sd_name,
     struct udev_device *sd_udev = NULL;
     int rc = LSM_ERR_OK;
     const char *wwn = NULL;
-    char sys_path[_MAX_SYSFS_BLK_PATH_STR_LEN];
 
     memset(vpd83, 0, _MAX_VPD83_NAA_ID_LEN);
-    snprintf(sys_path, _MAX_SYSFS_BLK_PATH_STR_LEN, _SYSFS_BLK_PATH_FORMAT,
-             sd_name);
 
     udev = udev_new();
     if (udev == NULL) {
@@ -362,7 +359,7 @@ static int _udev_vpd83_of_sd_name(char *err_msg, const char *sd_name,
         goto out;
     }
 
-    sd_udev = udev_device_new_from_syspath(udev, sys_path);
+    sd_udev = udev_device_new_from_subsystem_sysname(udev, "block", sd_name);
     if (sd_udev == NULL) {
         _lsm_err_msg_set(err_msg, "Provided disk not found");
         rc = LSM_ERR_NOT_FOUND_DISK;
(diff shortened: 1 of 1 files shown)


c_binding/lsm_scsi.c
```diff
@@ -253,30 +205,31 @@ static int _sysfs_vpd83_naa_of_sd_name(char *err_msg, const char *sd_name,
...
-    rc = _sysfs_read_file(err_msg, sysfs_path, buff, &read_size,
-                          _MAX_VPD83_PAGE_LEN);
+    sd_udev = udev_device_new_from_subsystem_sysname(udev, "block", sd_name);
+    if (sd_udev == NULL) {
+        rc = LSM_ERR_NOT_FOUND_DISK;
+        goto out;
+    }
 
-    if (rc != LSM_ERR_OK)
+    value = udev_device_get_sysattr_value(sd_udev, "device/vpd_pg83");
```
cathay4t: @mulkieran The last time I check udev sys property for 'device/vpd_pg83' is empty string. That's why read sysfs file directly.

My OS(archlinux) is still set it as empty:
```
[fge@Gris-Laptop ~]$ udevadm info -a /dev/sda|grep vpd
    ATTRS{vpd_pg80}==""
    ATTRS{vpd_pg83}==""udevadm info -a /dev/sda|grep vpd
```

cathay4t:
Thank you for your patches.

My intention is to eliminate the libudev in the PR #85 (not ready yet).
Could you kindly share your ideas there?

c_binding/lsm_scsi.c
mulkieran: @cathay4t So, it turns out that libudev is entirely inconsistent in its vpd_pg83 attribute handling. Sometimes it returns a value:

```
>>> dev = pyudev.Device.from_path(context, '/sys/devices/pci0000:80/0000:80:02.0/0000:81:00.0/host2/target2:0:9/2:0:9:0')
>>> dev.attributes.get('vpd_pg83')
'\r\x83'
```

sometimes it returns an empty string:

```
>>> dev
Device(u'/sys/devices/pci0000:00/0000:00:03.0/0000:03:00.0/host1/target1:0:0/1:0:0:0/block/sda')
>>> dev.attributes.get('device/vpd_pg83')
''
```

Both of the corresponding files contain some data.

In contexts where there really is nothing to read it returns NULL.

From what you've told me, I guess that is all dependent on the value of the bytes that it encounters when reading the file. So, probably libudev has a pretty pervasive problem when reading binary data from attributes.

My understanding is that udevadm shouldn't be showing vpd_pg83 values at alll, because --attribute-walk is supposed to only show those textual attributes that can be matched against udev rules. It shows the value because it is incorrectly read and therefore interpreted as text.

I think I'll file a broader bug against libudev about this inconsistancy in reading non-textual sysfs attributes, and see what happens.

But, I guess that what we may know at this point is that libudev's reading of all non-text attributes is likely to be unreliable.

mulkieran:
Thanks for looking. ASFAICT this PR is pretty well obsoleted already. Feel free to close at any time.

c_binding/lsm_scsi.c
cathay4t: In case you are curious, the reason inconsistent is caused libudev ATA endian problem:

The vpd_pg83 is starting with 00 83(big-edian).
That's why you got SAS disk with empty string while SATA disk with 0x83.

cathay4t:
Closing this PR and will continue the discuss in PR #85

PR #94, opened 2016-02-23
plugin/hpsa.py: Workaround for two attributes that the Dynamic Smart …

joehandzik:
…Array has alternate entries for in hpssacli output


joehandzik:
For reference, since this is vendor-specific:

*Regression tested against DL380 Gen9 with H240ar in HBA mode
*Tested for fix validation against Apollo 4520 with boot controller as B140i and data controllers H240 in HBA mode and H241 in HBA mode

tasleson:
Thanks @joehandzik !

joehandzik:
@tasleson You're welcome!

Issue #93, opened 2016-02-23
'sudo make check' fails with "Use of uninitialized value $class_line", 'make check' passes


joehandzik:
Hey guys,

Was really freaked out when my unit tests were failing for a very small commit, so I pulled master and ran again. I get a very quick 'make check' failure (running as root, so 'sudo make check'), with this error message:

Use of uninitialized value $class_line in pattern match (m//) at /home/jhandz/lsm_debug_unit_testing/tools/utility/check_const.pl line 283.

Pilot error, or does something need to be fixed? Whatever changed, it changed within the last 44 commits, that was the delta between my fork and libstoragemgmt's master.

EDIT: 'make check' passes without issue.

tasleson:
@joehandzik I just did:

``` bash
$ git pull && git checkout master && git clean -fxd && ./autogen.sh && \
./configure --without-rest-api && make check
```

worked for me.  

I never run any of these checks as root.

joehandzik:
You are right! This passes when not run using 'sudo'. Previously, sudo had always worked...lemme edit the title to make the issue clear, and I'll leave it up to you guys if it's an issue worth fixing.

joehandzik:
Even 'sudo -E' fails, so it's not an environment variable issue as far as I can tell, even though the error would lead me to believe that is the issue...

cathay4t:
@joehandzik @tasleson This issue has been fixed in PR https://github.com/libstorage/libstoragemgmt/pull/92

tasleson:
PR merged, closing.

PR #95, opened 2016-02-26
Fix 'sudo make check'

cathay4t:
* Fix error when run 'make check' as root user. Issue #93 
 * Fix constants check.


tasleson:
Looks good and tests good too!  Thanks Gris!

PR #96, opened 2016-03-02
Use pyudev to obtain sysfs attributes.

mulkieran:
Signed-off-by: mulhern <amulhern@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
         - python-dev
         - valgrind
         - chrpath
+        - python-pyudev
 
 compiler: gcc
 
--- a/configure.ac
+++ b/configure.ac
@@ -183,6 +183,7 @@ AM_PATH_PYTHON([2.6], [], AC_MSG_ERROR([Python interpreter 2.6 or 2.7 required])
 AC_PYTHON_MODULE([pywbem], [Required])
 AC_PYTHON_MODULE([M2Crypto], [Required])
 AC_PYTHON_MODULE([argparse], [Required])
+AC_PYTHON_MODULE([pyudev], [Required])
 PKG_CHECK_MODULES([LIBUDEV], [libudev])
 PKG_CHECK_MODULES([PYTHON2], [python2], [], [not_found_py_pkg=yes])
 
--- a/packaging/libstoragemgmt.spec.in
+++ b/packaging/libstoragemgmt.spec.in
@@ -256,6 +256,7 @@ MegaRAID storage management via storcli.
 Summary:        Files for HP SmartArray support for %{libstoragemgmt}
 Group:          System Environment/Libraries
 Requires:       %{libstoragemgmt}%{?_isa} = %{version}-%{release}
+Requires:       python-pyudev
 BuildArch:      noarch
 
 %description    -n %{libstoragemgmt}-hpsa-plugin
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -467,26 +467,23 @@ class SmartArray(IPlugin):
     @staticmethod
     def _hp_ld_to_lsm_vol(hp_ld, pool_id, sys_id, ctrl_num, array_num,
                           hp_ld_name):
+        """
+        raises DeviceNotFoundError
+        """
         ld_num = hp_ld_name[len("Logical Drive: "):]
         vpd83 = hp_ld['Unique Identifier'].lower()
         # No document or command output indicate block size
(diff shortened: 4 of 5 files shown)


mulkieran:
Another attempt to introduce pyudev dependency in libstoragemgmt.

The PR is incomplete, I didn't want to go any further with it unless it was at least a possibility.

Also, I'm not sure that using fallback values is really the right choice if the device doesn't exist at all, but that possibility didn't need to be addressed in the original source, so I do not know what is best.

joehandzik:
It's certainly cleaner, no arguments from me. If @cathay4t and @tasleson accept this, I'll rework mine the same way.

joehandzik:
@mulkieran I can't think of a reason why hpssacli would expose the volume but the host wouldn't have a sysfs path to it. Honestly, if that sort of thing is happening it doesn't matter (from my perspective) what values are present in any attributes, because we're already in crazy-town.

mulkieran:
@joehandzik It ought to be unlikely. I do a lot of CI with pyudev with tests drawn from the existing environment and I have seen it happen where the test fails because the device being tested disappeared. So I see these kinds of things more than most people.

joehandzik:
@mulkieran I suppose it's fair that the device could disappear after the python object has been instantiated...not so unusual in my experience either. Any ideas for better error detail?

cathay4t:
@mulkieran Thanks for the PR. Please add `python-pyudev` package into `.travis.yml`.
Meanwhile, could you provide test results on old system like RHEL 6?

mulkieran:
@cathay4t .travis.yml done. I'm not sure how I would go about getting satisfactory test for RHEL6, you'll have to advise me on that.

cathay4t:
@mulkieran Let me do that test by reserving a hpsa server in RHEL 6 from our internal beaker lab.

cathay4t:
@mulkieran Please update RPM spec file for dependency also. Thanks!

mulkieran:
@cathay4t spec file updated. My original code had a major bug that should definitely have raised an exception if it had been executed. I guess the CI tests aren't able to test the actual functionality of the plugins?

tasleson:
@mulkieran The unit tests only test the code against the simulators and the nstor and targetd boxes we have.  We can certainly add more boxes, but we need the hardware and CI service running to utilize, something we have been meaning to do.

mulkieran:
@tasleson Thanks, that makes sense. Is there any code in the repo for those two plugin tests? I'm not finding it, but maybe I don't know how to look.

tasleson:
@mulkieran We have test/plugin_test.py which runs against plugins.  You should be able to run that on a system that has the hardware needed to exercise the plugin(s).

tasleson:
@mulkieran Any update on running the test on actual hardware?

joehandzik:
@tasleson @mulkieran Seems like maybe this should be me, what with me having ten ProLiant servers of my own in a rack 30 feet away from me and all...let me schedule some cycles to do the testing this week. Sound fair?

tasleson:
@joehandzik Sounds fair to me, Thanks!

mulkieran:
@joehandzik Thanks!

joehandzik:
'lsmcli -lv' gives me the right size answer, which is based on two of the attributes you're pulling out of sys. I call it good!

tasleson:
@joehandzik Would it be possible to run the `test/plugin_test.py` ?

joehandzik:
@tasleson T'was the first thing I did, it passed all 'make check' tests. I did only test Rhel7 though, so good call on testing Rhel6.

joehandzik:
@tasleson Passed, ran 42 tests in ~70 seconds.

tasleson:
@joehandzik `make check` only does tests against the simulators, C & Python.  To test against a piece of hardware we use the `test/plugin_test.py`.  You would need to make sure that a pool exists before hand to allow the test the ability to create things.  I've not actually run the test myself against hardware, but @cathay4t has.

joehandzik:
@tasleson Yeah, I realized my error a few minutes ago. See my updated comments above.

tasleson:
@joehandzik great, thanks!

cathay4t:
@mulkieran Please update `configure.ac` to check `pyudev` module. This line might helps:
```
AC_PYTHON_MODULE([pyudev], [Required])
```

Besides that, RHEL 6 test on hpsa works well.

tasleson:
@mulkieran @joehandzik @cathay4t Thanks everyone for the improvement!

PR #101, opened 2016-03-30
hpsa.py: Add SYS_MODE_GET to capabilities list

joehandzik:
The Smart Array plugin was missing the SYS_MODE_GET Capability, even though it clearly supports the feature and is collecting the data.


tasleson:
Thanks @joehandzik !

PR #102, opened 2016-03-30
libstoragemgmt: Add read_cache_pct system property

joehandzik:

### DIFF
--- a/c_binding/lsm_datatypes.cpp
+++ b/c_binding/lsm_datatypes.cpp
@@ -732,6 +732,7 @@ lsm_system *lsm_system_record_alloc(const char *id, const char *name,
         rc->status_info = strdup(status_info);
         rc->fw_version = NULL;
         rc->mode = LSM_SYSTEM_MODE_NO_SUPPORT;
+        rc->read_cache_pct = LSM_SYSTEM_CACHE_PCT_NO_SUPPORT;
 
         if (plugin_data) {
             rc->plugin_data = strdup(plugin_data);
@@ -787,6 +788,7 @@ lsm_system *lsm_system_record_copy(lsm_system * s)
         rc = lsm_system_record_alloc(s->id, s->name, s->status, s->status_info,
                                      s->plugin_data);
         rc->mode = s->mode;
+	rc->read_cache_pct = s->read_cache_pct;
 
         if ((s->fw_version != NULL) &&
             (lsm_system_fw_version_set(rc, s->fw_version) != LSM_ERR_OK)) {
@@ -869,6 +871,30 @@ int lsm_system_mode_get(lsm_system *s, lsm_system_mode_type *mode)
     *mode = s->mode;
 
     if (s->mode != LSM_SYSTEM_MODE_NO_SUPPORT)
+        return LSM_ERR_OK;
+    else
+        return LSM_ERR_NO_SUPPORT;
+}
+
+int lsm_system_read_cache_pct_set(lsm_system *sys, int read_pct)
+{
+    if ((sys == NULL) || (! LSM_IS_SYSTEM(sys)) ||
+        (read_pct == LSM_SYSTEM_CACHE_PCT_NO_SUPPORT))
+        return LSM_ERR_INVALID_ARGUMENT;
+
+    sys->read_cache_pct = read_pct;
+
+    return LSM_ERR_OK;
+}
+
+int lsm_system_read_cache_pct_get(lsm_system *s, int *read_pct)
+{
+    if ((s == NULL) || (read_pct == NULL) || (! LSM_IS_SYSTEM(s)))
+        return LSM_ERR_INVALID_ARGUMENT;
+
+    *read_pct = s->read_cache_pct;
+
+    if (s->read_cache_pct != LSM_SYSTEM_CACHE_PCT_NO_SUPPORT)
         return LSM_ERR_OK;
     else
         return LSM_ERR_NO_SUPPORT;
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -155,6 +155,8 @@ typedef enum {
     LSM_CAP_EXPORT_CUSTOM_PATH = 124,
     /**^ Plug-in allows user to define custome export path */
 
+    LSM_CAP_SYS_READ_CACHE_PCT_GET = 159,
+    /**^ Plug-in allows user to retrieve system read cache percentage */
     LSM_CAP_SYS_FW_VERSION_GET = 160,
     /**^ Plug-in allows user to retrieve storage firmware version */
     LSM_CAP_SYS_MODE_GET = 161,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1315,6 +1315,16 @@ lsm_system LSM_DLL_EXPORT *lsm_system_record_alloc(const char *id,
                                                    const char *plugin_data);
 
 /**
+ * New in version 1.3. Set read cache percentage.
+ * @param[in] id            Id
+ * @param[in] sys           System to update.
+ * @param[in] read_pct      Read cache percentage.
+ * @return LSM_ERR_OK or LSM_ERR_INVALID_ARGUMENT.
+ */
+int LSM_DLL_EXPORT lsm_system_read_cache_pct_set(lsm_system *sys,
+                                                 int read_pct);
+
+/**
  * New in version 1.3. Set firmware version.
  * @param[in] id            Id
  * @param[in] sys           System to update.
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_systems.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_systems.h
@@ -70,6 +70,16 @@ const char LSM_DLL_EXPORT *lsm_system_id_get(lsm_system *s);
 const char LSM_DLL_EXPORT *lsm_system_name_get(lsm_system *s);
 
 /**
+ * New in version 1.3. Retrieves read cache percentage of the specified system.
+ * @param       s      		System to retrieve read cache percentage for.
+ * @param[out]  read_pct  	Read cache percentage.
+ * @return LSM_ERR_OK on success, or LSM_ERR_NO_SUPPORT, or
+ *         LSM_ERR_INVALID_ARGUMENT.
+ */
+int LSM_DLL_EXPORT lsm_system_read_cache_pct_get(lsm_system *s,
+                                                 int *read_pct);
+
+/**
  * Retrieve the status for the system.
  * @param s     System to retrieve status for
  * @return System status which is a bit sensitive field, returns UINT32_MAX on
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
@@ -337,6 +337,9 @@ typedef enum {
     LSM_SYSTEM_MODE_HBA = 2,
 } lsm_system_mode_type;
 
+#define LSM_SYSTEM_CACHE_PCT_NO_SUPPORT             -2
+#define LSM_SYSTEM_CACHE_PCT_UNKNOWN                -1
+
 #ifdef  __cplusplus
 }
 #endif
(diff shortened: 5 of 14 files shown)


tasleson:
@joehandzik When I think ratio I think numeric, but I see you are using a string?  Please elaborate on this and a little back story about what this feature is used for too as I'm not familiar with it, thanks!

tasleson:
@joehandzik Thanks for the information.  So you can use 10% read and 20% for write and 70% unused?  I'm thinking you would want the ability to disable the entire thing, use it all for read caching or a mixture of read & write or all write cache?  With that use case I could see using a signed integer for % read cache where:
```
-1     == All caching disabled,
0      == All write cache
1-99   == Percent read, remaining write
100    == All read cache

```

IMHO, If we need 2 knobs to twist, then having a pair of integers would be better, especially for C API users where string manipulation is a pain, plus no stings to allocate and free.

joehandzik:
@tasleson No, and there wouldn't really be any reason to do that (if its not used for volume caching, it would just be unused entirely). 

Users effectively "disable" the controller cache by never configuring a volume to use it.

You're right to say that we don't need two separate knobs (if you choose 30% for your read cache, the write cache will be at 70%). I like your plan for configuring the cache percentage. Do you still think I need to report the cache values as separate integers? I could do it if you think that's better.

tasleson:
No, my preference would be the 1 integer value.  With your clarification, I see no reason to have 2.

joehandzik:
@tasleson Updated to switch to an integer-based implementation. Lemme know if you dislike the fact that I don't have any mechanism to communicate that a plugin doesn't support the operation, or if there's a way you'd like that documented. I interpret the -1 more as "I don't know what the read cache percentage is".

tasleson:
@joehandzik I'm a little confused by your concern.  A plugin can convey that it does or doesn't support the operation with the capability.  Furthermore it is entirely possible for a plugin to support something, but that a particular feature isn't available for this specific instance of hardware, which should be conveyed with the capability too.  Additionally, the get & set have a return code, which can return not supported if the user does happen to call it without checking capability first.  Having the get return no error with a -1 read percentage seems like a suitable answer to I don't know.

What are the possible reasons that you are unable to retrieve a value?
* No volumes are configured to use the cache?
* ...

For this specific operation we could certainly define a number of negative values to return in the percent field if we really need to convey more information too.

joehandzik:
Fair point! I'm just used to how @cathay4t has been writing this sort of feature, and it generally involves an extra mechanism to communicate that the feature isn't supported via the get & set return code. I haven't done any of that here, I'm just messaging support via the plugin capability list.

Possible reasons:

1. No volumes are configured to use the cache (as you said)
2. Controller doesn't actually have user-accessible cache

That's it, as far as I know.

There aren't any configs where the value could only be read on Smart Array. This specific pull only allows for reading the cache value. I'll put together a separate feature for updating the cache percentages when this one is agreed on. Keeping the features split would allow for a system to support read-only.

tasleson:
@joehandzik @cathay4t Can you elaborate on this "extra mechanism" that Gris is doing, maybe point to an example in the code?

Seems like #2 would be handled by not returning the fact that it has this capability.  For example, the SMI-S plugin is generic across many different supported arrays.  The capability returned is for that specific hardware instance.

Also for the `set` discussion, I was getting confused about just setting the value (which a plugin does) on the object and allowing the user to change the value for the hardware, which does indeed require some addition code, sorry about that.

FYI: Gris should be back tomorrow :-)

joehandzik:
@tasleson Bits of code like this: https://github.com/libstorage/libstoragemgmt/blob/master/python_binding/lsm/_data.py#L406
and this: https://github.com/libstorage/libstoragemgmt/blob/master/c_binding/lsm_convert.cpp#L319

There are several other examples across both C and python bindings if you want me to pick out more. But the gist is that he takes hints from values in various locations and interprets them as meaning "this field is not supported on the hardware we're talking to right now". I'm not doing any of that, mostly because the read cache percentage isn't an enum right now. I suppose I could go through and make -2 or something the "not supported" value if we decide it's necessary. Just lemme know what you guys want.

Well, I haven't seen capabilities returned differently for different hardware models of Smart Array in the hpsa plugin. While we could do that, it'd be kind of a headache because you'd essentially need to create a list of controllers that don't have cache as a sort of blacklist. I'd prefer, for simplicity's sake, to just say that we don't know what the cache percentage is for all those controllers.

Glad to have Gris back! He hasn't stayed completely away though, I've watched his hardware LED branch move. :)

cathay4t:
I guess the "extra mechanism" Joe means might be raise `NO_SUPPORT` for this property if plugin doesn't set it in object initiation. Example(Based on Tony's suggestion):
```
-2     == No support
-1     == All caching disabled,
0      == All write cache
1-99   == Percent read, remaining write
100    == All read cache

```

I am afraid HPSA will be the only plugin support this feature:
 * MegaRAID: does not expose this info.
 * SMI-S: no section for cache management. But EMC has their extention to support this, but I don't go
   through that mess again.
 * ONTAP: expose no cache setting yet.
 * Nstor: no idea.
 * Targetd: no such feature.

Yeah, the SES LED library function is drafted, please review them as well. :)

joehandzik:
@cathay4t Yeah, that's what I was attempting to describe. Do you guys want that, or can I leave the pull as-is?

That's kind of a bummer that nobody else supports it. I assume it's still an ok feature to add? I'd be kinda bummed if we start keeping features out of lsm just because nobody else supports them (or doesn't support them yet, anyway). 

Anyway, lemme know if you guys want other changes. Otherwise I'll assume this is good. I'm targeting several features beyond the Ceph requirements we've already talked about for the next month, in time for the May 1st release.

cathay4t:
Even hpsa will be the only plugin support this feature, I would like to give my  +1 for this.

joehandzik:
@cathay4t Thanks! Any required updates to the pull then, or can it be merged?

tasleson:
@joehandzik After looking at this some more I think I would like to see adding constants for -2 == NOT_SUPPORTED, -1 == UNKNOWN_OR_DISABLED.  This way we initialize the the value to -2 as the library wide default, so all other plugins when called will get not supported.  In the case of this plugin, if you cannot determine the value, set it to -1, otherwise the actual value.  Of course you need to add some checks in the getters and translate to correct error code.

Does this sound reasonable?

joehandzik:
@tasleson Yup, works for me! I'll try to get that done today.

joehandzik:
@tasleson So, I made the changes you suggested and rebased with master to remove the branch conflicts, but it would appear that Github is working through a backlog (as per their status page), so the updates aren't showing up yet. Lemme know what you think sometime after it all updates!

tasleson:
@joehandzik  Will do, looks like today has been a bad day for github WRT uptime.

c_binding/lsm_datatypes.cpp
```diff
@@ -874,6 +876,28 @@ int lsm_system_mode_get(lsm_system *s, lsm_system_mode_type *mode)
...
+}
+
+int lsm_system_read_cache_pct_get(lsm_system *s, int *read_pct)
+{
+    if ((s == NULL) || (read_pct == NULL) || (! LSM_IS_SYSTEM(s)))
+        return LSM_ERR_INVALID_ARGUMENT;
+
+    if (s->read_cache_pct != LSM_SYSTEM_CACHE_PCT_NO_SUPPORT)
+        *read_pct = s->read_cache_pct;
+
```
tasleson: When `s->read_cache_pct ==  LSM_SYSTEM_CACHE_PCT_NO_SUPPORT`  I was expecting that we would be returning `LSM_ERR_NO_SUPPORT` like we do for python?

joehandzik: We could do that. We'll need to change the behavior for fw_version and mode set too, that's what I based my implementation on.

Joe

On Apr 6, 2016, at 9:08 AM, Tony Asleson <notifications@github.com<mailto:notifications@github.com>> wrote:


In c_binding/lsm_datatypes.cpp<https://github.com/libstorage/libstoragemgmt/pull/102#discussion_r58710967>:

> +        (read_pct == LSM_SYSTEM_CACHE_PCT_NO_SUPPORT))
> +        return LSM_ERR_INVALID_ARGUMENT;
> +
> +    sys->read_cache_pct = read_pct;
> +
> +    return LSM_ERR_OK;
> +}
> +
> +int lsm_system_read_cache_pct_get(lsm_system *s, int *read_pct)
> +{
> +    if ((s == NULL) || (read_pct == NULL) || (! LSM_IS_SYSTEM(s)))
> +        return LSM_ERR_INVALID_ARGUMENT;
> +
> +    if (s->read_cache_pct != LSM_SYSTEM_CACHE_PCT_NO_SUPPORT)
> +        *read_pct = s->read_cache_pct;
> +


When s->read_cache_pct == LSM_SYSTEM_CACHE_PCT_NO_SUPPORT I was expecting that we would be returning LSM_ERR_NO_SUPPORT like we do for python?

-

tasleson: On 04/06/2016 09:16 AM, Joe Handzik wrote:
> We could do that. We'll need to change the behavior for fw_version
> and mode set too, that's what I based my implementation on.

I'm confused by this statement as I see this for the implementation of
the 'get' operations for fw version and mode, both of which will return
LSM_ERR_NO_SUPPORT?

int lsm_system_fw_version_get(lsm_system *s, const char **fw_ver)
{
    if ((s == NULL) || (fw_ver == NULL) || (! LSM_IS_SYSTEM(s)))
        return LSM_ERR_INVALID_ARGUMENT;

    *fw_ver = s->fw_version;

    if (s->fw_version[0] != '\0' )
        return LSM_ERR_OK;
    else
        return LSM_ERR_NO_SUPPORT;
}

int lsm_system_mode_get(lsm_system *s, lsm_system_mode_type *mode)
{
    if ((s == NULL) || (mode == NULL) || (! LSM_IS_SYSTEM(s)))
        return LSM_ERR_INVALID_ARGUMENT;

    *mode = s->mode;

    if (s->mode != LSM_SYSTEM_MODE_NO_SUPPORT)
        return LSM_ERR_OK;
    else
        return LSM_ERR_NO_SUPPORT;
}

Regards,
Tony

joehandzik: Ah, I misunderstood which read_pct check you were talking about (both showed up in the code snapshot I saw in email, and I chose the first check rather than the second).

Will fix as you suggested for the gets.

Joe

On Apr 6, 2016, at 9:35 AM, Tony Asleson <notifications@github.com<mailto:notifications@github.com>> wrote:


In c_binding/lsm_datatypes.cpp<https://github.com/libstorage/libstoragemgmt/pull/102#discussion_r58715718>:

> +        (read_pct == LSM_SYSTEM_CACHE_PCT_NO_SUPPORT))
> +        return LSM_ERR_INVALID_ARGUMENT;
> +
> +    sys->read_cache_pct = read_pct;
> +
> +    return LSM_ERR_OK;
> +}
> +
> +int lsm_system_read_cache_pct_get(lsm_system *s, int *read_pct)
> +{
> +    if ((s == NULL) || (read_pct == NULL) || (! LSM_IS_SYSTEM(s)))
> +        return LSM_ERR_INVALID_ARGUMENT;
> +
> +    if (s->read_cache_pct != LSM_SYSTEM_CACHE_PCT_NO_SUPPORT)
> +        *read_pct = s->read_cache_pct;
> +


On 04/06/2016 09:16 AM, Joe Handzik wrote: We could do that. We'll need to change the behavior for fw_version and mode set too, that's what I based my implementation on.
I'm confused by this statement as I see this for the implementation of the 'get' operations for fw version and mode, both of which will return LSM_ERR_NO_SUPPORT? int lsm_system_fw_version_get(lsm_system *s, const char **fw_ver) { if ((s == NULL) || (fw_ver == NULL) || (! LSM_IS_SYSTEM(s))) return LSM_ERR_INVALID_ARGUMENT; *fw_ver = s->fw_version; if (s->fw_version[0] != '\0' ) return LSM_ERR_OK; else return LSM_ERR_NO_SUPPORT; } int lsm_system_mode_get(lsm_system *s, lsm_system_mode_type *mode) { if ((s == NULL) || (mode == NULL) || (! LSM_IS_SYSTEM(s))) return LSM_ERR_INVALID_ARGUMENT; *mode = s->mode; if (s->mode != LSM_SYSTEM_MODE_NO_SUPPORT) return LSM_ERR_OK; else return LSM_ERR_NO_SUPPORT; } Regards, Tony

-

joehandzik:
@tasleson Addressed the issue you pointed out. Thanks!

tasleson:
Thanks @joehandzik !

PR #107, opened 2016-04-06
[PATCH 0/2] Fixes for HP SmartArray plugin.

cathay4t:
*  HPSA Plugin: Fix disk querying in RAID mode.
*  HPSA plugin: Fix RAID volume creating.

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -386,7 +386,7 @@ class SmartArray(IPlugin):
 
         for ctrl_name in ctrl_all_show.keys():
             ctrl_data = ctrl_all_show[ctrl_name]
-            try: 
+            try:
                 sys_id = ctrl_data['Serial Number']
             except KeyError:
                 #Dynamic Smart Array does not expose a serial number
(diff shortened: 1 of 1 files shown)


plugin/hpsa/hpsa.py
```diff
@@ -755,7 +761,7 @@ def volume_raid_create(self, name, raid_type, disks, strip_size,
             cmds.append("ss=%d" % int(strip_size / 1024))
 
         try:
-            self._sacli_exec(cmds, flag_convert=False)
```
joehandzik: I still really, really don't like this change as default behavior. You should not need to force create a volume. 

From your email a few weeks ago you made it seem like this issue only happened when you tried to change the initial volume setting. Did you encounter behavior where a force create is required at initial deployment?

joehandzik: Ok, answered my own question by reading your notes in the individual commit for this change. Overprovisioning is actually a really important feature for many customers. There are workflows that involve deleting a volume and then recreating it with overprovisioning to have the Smart Array send TRIM to the disk, improving SSD performance...we'll want to be sure that a force option causes TRIM to be sent...

tasleson:
Commit message for second commit:

`Plugin will hand when creating volume using SSD disk.`
Typo `hand -> hang`

Otherwise I think it looks good.

cathay4t:
Commit message amended. Thanks, Tony!

Let's wait feedback from @joehandzik .

joehandzik:
@cathay4t @tasleson Just a quick note to let you know that I'm still working this one. I need to take a look at exactly what that overprovisioning message looks like...depending on the phrasing, it's either expected behavior or a bug. If it's not a bug, then your 'forced' option works fine.

joehandzik:
@cathay4t I just tested this out on my box, and confirmed that the phrasing falls in the "expected behavior" bucket. You're safe to merge this in from my perspective.

tasleson:
@cathay4t @joehandzik , Thanks everyone!

PR #110, opened 2016-04-08
HPE Smart Array: Handle the full variety of ways that system cache ca…

joehandzik:
…n be reported by hpssacli

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -400,10 +400,13 @@ class SmartArray(IPlugin):
             except KeyError:
                 #Dynamic Smart Array does not expose a firmware version
                 fw_ver = "%s" % ctrl_data['RAID Stack Version']
-            try:
+            if 'Cache Ratio' in ctrl_data:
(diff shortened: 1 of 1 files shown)


joehandzik:
Our lovely cli utility supports multiple terms for the controller cache. Apologies!

joehandzik:
nstor strikes again!

tasleson:
@joehandzik I looked at this nstor failure some more today, it's the same issue we've hit before.  I will try and reach out to nexenta and see if we can get to root cause.

cathay4t:
The nstor error is caused by empty VPD83 issue reported by plugin_test.
If we failed to find a way to get nstor vpd83, we could update `plugin_test.py` to skip vpd83 test on nstor URI and mark it as known issue.

joehandzik:
Any thoughts on this guy? Or do you want to wait on it until I can rebase in a fix for the nstor failures?

tasleson:
@joehandzik Thanks for the fix!

tasleson:
@joehandzik We are going to change nstor from failure to known issue here shortly.  Disregard any failures when the error is missing vpd83.

PR #113, opened 2016-04-12
plugin_test.py: Only write messages to stderr

tasleson:
The automated tests that run in the lab use stderr for generic
messaging and stdout to report execution timings.  As the unit
test prints to stderr this all works fine, except when I added
some print calls for warnings on the cleanup change in
a542eb2b04b4f8d07a2f320d965ce30fe8bfd1fe.  This caused the
automated tests to fail as we have invalid text in stdout which
was causing the yaml parser to fail.

Note(s):
- 2 of the arrays are returning errors during clean-up
- This change tested on all arrays in the lab

### DIFF
--- a/test/plugin_test.py
+++ b/test/plugin_test.py
@@ -30,7 +30,6 @@ import argparse
 import collections
 import atexit
 import sys
-import re
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


cathay4t:
Tested. Merged.

PR #114, opened 2016-04-13
libstoragemgmt: Add read cache percentage update feature.

joehandzik:
This feature allows a user to update the read cache percentage of a storage system, when supported. For many implementations, this will implicitly change the write cache percentage to be the remaining available percentage after the read cache percentage is updated.

Tested on a P840ar in an Apollo 4200.


cathay4t:
@joehandzik Could you change method name from `lsm_xxx_update()` to `lsm_xxx_set()`? Just want to be consistent on function naming schme.

joehandzik:
So, the reason why I didn't do that is because there's already a lsm_read_cache_pct_set here: https://github.com/libstorage/libstoragemgmt/blob/9e50995ae4e3da66ffe1a240870e52b6dbc30dd5/c_binding/lsm_datatypes.cpp#L879

Lemme know what go forward plan you suggest. Thanks!

tasleson:
@joehandzik @cathay4t  This is a good question.  We basically have getters and setters on object types, which are using the `_get` and `_set` verbiage.  In this case Joe is trying to change the value of one of the system values, not just test the value in the object.

joehandzik:
@tasleson Right. So far, I think @cathay4t is right that the functions have been named *_set, but this case required a different naming scheme (IMO). I'm fine with choosing a different naming scheme or rethinking the current setter naming scheme for recent feature adds (controller mode, firmware version, read cache percentage, etc).

cathay4t:
How about we change plugin interface naming scheme to `lsm_plug_xxx_set()`?
Even we cannot change function name defined in 1.2 and before, we still change change function name introduced in 1.3.

joehandzik:
@cathay4t That sounds fine, I'll give that a shot today too. Thanks!

joehandzik:
@cathay4t lsm_plug_xxx_set doesn't work in this case EITHER, because of this: https://github.com/joehandzik/libstoragemgmt/blob/wip-read-cache-set/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h#L1017

I went with lsm_internal_system_read_cache_pct_set, it's not much longer and probably a bit more accurate regarding the purpose of the function. Hope that works for you!

tasleson:
@joehandzik @cathay4t IMHO I think we should not change the naming convention for established things.  If we do this it just makes using the library more confusing.  Right or wrong, I think we should stay consistently the same.  I'm actually thinking we should adopt some other term to denote propagating a change, perhaps going back to Joe`s original name `_update` and then stick to using that for other things that change?  We need a way to denote changing the state of the object vs. changing the state of the system.

Otherwise I have a couple of other ideas

1. Changing how we update things.  For example if a client API user wanted to update one or more things, using this existing approach they have to round trip for each item.  If we allowed the client API user to use the set methods on an object and then pass that object into the framework to update, we could update the entire state of the object in one round trip.  However, this introduces difficulty if one of the individual updates in the object fail.  How to report which one(s) and what is the state of the update, all previous or all new except one(s) that failed.

2. If we could devise a way to distinguish between a client `_set` vs. a plugin `_set`.  We could then change the implementation of set so that the client set actually makes a round trip over the IPC to update the state of the system, whereas a plugin set just updates the state of the object representing the  hardware state.

joehandzik:
@tasleson 

1: This could be relevant for Redfish-style updating in the future. Lemme take a look at how failures when multiple updates are attempted are messaged there. I honestly don't know.
2: I was partial to my _update naming schema for the client set, personally. :) I agree that there's a logical difference between the two that should be apparent via the name.

joehandzik:
I still need to find cycles to understand what Redfish does in this case, but for the time being...what should I change this to in order to get the code in place for release 1.3? I can revert to the _update semantics if we're all in agreement on that.

tasleson:
@joehandzik I think using `update` would be fine, but lets get consensus with Gris too.

tasleson:
`_update` has consensus, we are going with that.

cathay4t:
+1 for `_update`

joehandzik:
@tasleson @cathay4t Updated to _update. Lemme know if something else is needed!

tasleson:
Thanks @joehandzik

PR #115, opened 2016-04-14
libstoragemgmt: Add volume status field to volumes

joehandzik:
This feature allows a user to see the health status of their volumes. Statuses that are handled explicitly: OK, DEGRADED, RECOVERING, and FAILED. Statuses that don't fall into those categories fall into the 'OTHER' status.

Tested on a P840ar in an Apollo 4200.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_capabilities.h
@@ -165,6 +165,8 @@ typedef enum {
     /**^ Plug-in allows user to retrieve disk location */
     LSM_CAP_VOLUME_LED = 171,
     /**^ Plugin allows user to set and clear volume LEDs */
+    LSM_CAP_VOLUME_STATUS = 172,
+    /**^ Plugin allows user to retrieve volume health status */
     LSM_CAP_POOLS_QUICK_SEARCH = 210,
     /**^ Seach occurs on array */
     LSM_CAP_VOLUMES_QUICK_SEARCH = 211,
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_plug_interface.h
@@ -1353,6 +1353,14 @@ int LSM_DLL_EXPORT lsm_system_mode_set(lsm_system *sys,
 const char LSM_DLL_EXPORT *lsm_system_plugin_data_get(lsm_system *s);
 
 /**
+ * New in version 1.3. Set volume status.
+ * @param[in] v             Volume to update.
+ * @param[in] status        Volume status 'lsm_volume_status_type'.
+ * @return LSM_ERR_OK or LSM_ERR_INVALID_ARGUMENT.
+ */
+int LSM_DLL_EXPORT lsm_volume_status_set(lsm_volume *v,
+                                         lsm_volume_status_type status);
+/**
  * Allocates storage for Access_group array
  * @param size      Number of elements to store.
  * @return  NULL on error, else pointer to array for use.
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
@@ -179,6 +179,15 @@ typedef enum {
     /**^ Vendor specific RAID type */
 } lsm_volume_raid_type;
 
+    /** lsm_volume_status_type: Types for volume health */
+typedef enum {
+    LSM_VOLUME_STATUS_NO_SUPPORT = 0,
+    LSM_VOLUME_STATUS_OTHER = 1,
+    LSM_VOLUME_STATUS_OK = 2,
+    LSM_VOLUME_STATUS_DEGRADED = 3,
+    LSM_VOLUME_STATUS_RECONSTRUCTING = 4,
+    LSM_VOLUME_STATUS_ERROR = 5,
+} lsm_volume_status_type; 
 
     /**^ \enum lsm_pool_member_type Different types of Pool member*/
 typedef enum {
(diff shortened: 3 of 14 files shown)


cathay4t:
@joehandzik We remove `Volume.status` a while back considering it's just a duplication of `Pool.status`.
Any special reason user should not use `Pool.status` as volume are always been allocated from pool in the terminology of lsm.
Meanwhile, please rename these status to align with `Pool.status`:
 * `RECOVERING` -> `RECONSTRUCTING`
 * `FAILED` -> `ERROR`

joehandzik:
@cathay4t I'll update the status definitions tomorrow. 

A few reasons:

1. For lots of use cases, the entire pool will be consumed by a single volume. When that's the case, the pool APIs can be ignored entirely and the user can pull everything they need from the volume APIs. When that's the level of abstraction that most users will be dealing with, I'd rather have the health attached to the volume.
2. If there are multiple volumes provisioned from a single pool, you may want to monitor their health separately. I'm not sure that there's a sensible use case for creating different types of volumes off a single array, or even if it's possible. But if you can create a RAID 0 off part of a disk in an array and create a RAID 1 off other disks, the pool could be in a degraded state with one volume being totally healthy. 
3. I'm going to add another API method to reenable a failed volume (useful when hot plugging a RAID 0 volume, allows users to avoid a delete and recreate option). Since that's executed against the volume, it feels more natural to check the volume's health directly to figure out if it should be reenabled rather than checking the volume's pool's health.

joehandzik:
Ok, updated the definitions. Lemme know if we need to talk more about if this feature is necessary or not (from my perspective, it's pretty key).

tasleson:
@joehandzik @cathay4t 

I believe the expectation was that a pool was the collection of disks/devices that were combined in some type of raid configuration and that each volume carved from the pool would share the same redundancy.  Thus it didn't make sense to have health status on each volume.  As for your reasons Joe, number 1 doesn't seem to be very compelling to me.  Number 2 only makes sense if you have a solution that allows a user to take a collection of disks (pool) and allow the same disk to be used in different redundancy configurations (eg. half a disk in RAID5, half in RAID1) or just different raid levels in the same pool with different disks.  However, I can think of at least 1 solution that allows both of these scenarios, LVM & MD.

We took a number of things out of release 1, as we knew we could always add things back in, but never take them out.  I guess I'm not against adding this back, especially now when we consider the flexibility that LVM & MD provides.  It might be useful to compare this change with what we took out, to see how they compare.

joehandzik:
Ok, let's spend some time comparing next week then and make a final decision.

Number 1 may not be super compelling from a code/API usage standpoint, but with the Smart Array plugin the ability to avoid using the pool method kills at least one extra call to hpssacli. That could make a positive difference for solutions that are polling volumes/pools/disks over and over again.

cathay4t:
For user case 2, adding a property for non-supported feature/plugin seems not a good way to go.
For user case 1, user could always use pool.status and pool terminology should not be ignored as when something goes wrong, user still need `lsm.Client.pool_member_info()` to find out which disk information.
For user case 3, how about `lsm.Client.volume_disable()` and `lsm.Client.volume_enable()`? And add/removing disks should be pool action, not volume which is just a logical allocation from pool.

Another reason HPSA might need `Volume.status` is their pool is not persistent[1], but this should not be a reason here.

About saving extra call, after pool/volume ownership established, polling pool status actually save
extra call if that pool has multiple volumes.

Current volume status monitor workflow (which suits both DAS and SAN):

 * Find out which pool owns interested volume by checking its `lsm.Volume.id`.
 * Poll `lsm.Client.pools()` on a regular interval.
 * If `lsm.Pool.status != lsm.Pool.STATUS_OK`, invoke `lsm.Client.pool_member_info()` to
   find out its RAID member disks.
 * Invoke `lsm.Client.disks()` to find out which disks goes wrong.

Adding `lsm.Volume.status` could simply the work flow.

[1]: Pool ID might change if some volume(not the last created volume) got deleted. From LSM's view, it's a bug and only hpssacli could fix it.

joehandzik:
Lemme see if I can reply in order, @cathay4t...

Use case 3 suggestion: I thought about that, but from what I saw of the APIs this was more of an administrative enable/disable rather than what I felt like I was trying to accomplish. If you and @tasleson agree that I can overload the volume_enable API to include re-enabling volumes that have been automatically marked as failed by the underlying storage system, I'll do it that way. That minimizes code changes, and helps get the feature in more quickly.

Regarding saving extra call: You're right that in a multiple-volumes-to-one-pool use case, checking the pool is more efficient. I tend to focus on single-volume-to-one-pool use cases in scale-out configs, but agree on the value of the pool status.

Regarding current monitoring workflow: Also agree on this one. It *can* work, assuming the volume health directly translates into the pool's health in all use cases. It does for mine, but as @taslson mentioned, there are some solutions that can end up with volume health that doesn't mirror the pool health. 

I'm ok punting on this one, I suppose. We'll see how the user I'm targeting reacts (this may end up used in Ceph at some point too), the pool health is adequate for now. Though I would like input on how I should implement volume re-enable (overload volume_enable, or brand new feature PR?). Thanks guys!

tasleson:
@joehandzik Is your main concern what a user gets returned from the CLI or how the API works for the health status?  If it's the former we could entertain displaying the health status when you list the volumes to make it easier for the end user.

joehandzik:
How the health API works. I don't really develop around the CLI, I just add features to it for completeness and as a quick integration test path with hpssacli.

Joe

On Apr 19, 2016, at 4:06 PM, Tony Asleson <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> Is your main concern what a user gets returned from the CLI or how the API works for the health status? If it's the former we could entertain displaying the health status when you list the volumes to make it easier for the end user.

-

joehandzik:
Gonna close this one for now to reduce clutter. We can revisit if necessary.

PR #119, opened 2016-04-21
HPE Smart Array: Older controllers don't report mode

### LINKED ISSUE #118: Controller mode not present for older controllers on older firmware
On a P420i, a user ran into a situation where the latest hpssacli did not expose the controller mode. I'll fix in a PR today.

joehandzik:
Some older Smart Arrays and more recent Smart Arrays with old firmware will fail to report a controller mode. Assume RAID mode for all these devices, which is a fair assumption on Linux.

Fixes #118 


tasleson:
Thanks! @joehandzik

PR #122, opened 2016-04-21
HPE Smart Array: Add support for volume_enable to Smart Array

joehandzik:

Lemme know what you guys think about this one for my volume reenable requirement, @tasleson and @cathay4t. My only concern is if a user will know that they are able to reenable a volume, but I suppose the workflow would be:

-assess pool health, discover that a pool is failed
-assess disk health, figure out which disk needs to be replaced (likely via Port:Box:Bay location)
-physically replace the device
-assess disk health again, if disks are all healthy but pool is still marked as unhealthy, attempt to reenable the volume provisioned from that pool

Thanks!

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -349,6 +349,7 @@ class SmartArray(IPlugin):
         cap.set(Capabilities.POOL_MEMBER_INFO)
         cap.set(Capabilities.VOLUME_RAID_CREATE)
         cap.set(Capabilities.VOLUME_DELETE)
+        cap.set(Capabilities.VOLUME_ENABLE)
         cap.set(Capabilities.SYS_FW_VERSION_GET)
         cap.set(Capabilities.SYS_MODE_GET)
         cap.set(Capabilities.SYS_READ_CACHE_PCT_GET)
@@ -506,12 +507,16 @@ class SmartArray(IPlugin):
             block_size = 512
             num_of_blocks = int(_hp_size_to_lsm(hp_ld['Size']) / block_size)
 
+        if 'Failed' in hp_ld['Status']:
+            admin_status = Volume.ADMIN_STATE_DISABLED
+        else:
+            admin_status = Volume.ADMIN_STATE_ENABLED
         plugin_data = "%s:%s:%s" % (ctrl_num, array_num, ld_num)
 
         # HP SmartArray does not allow disabling volume.
         return Volume(
             vpd83, vol_name, vpd83, block_size, num_of_blocks,
-            Volume.ADMIN_STATE_ENABLED, sys_id, pool_id, plugin_data)
+            admin_status, sys_id, pool_id, plugin_data)
 
     @_handle_errors
     def volumes(self, search_key=None, search_value=None,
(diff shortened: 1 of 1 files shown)


tasleson:
@joehandzik @cathay4t The only question I have about this change is if we are implementing this method with the same behavior as intended?  IIRC I thought this was more of an administrative `enable` and `disable`, but maybe I'm mistaken.

tasleson:
@joehandzik Sorry, I think I must have messed up the commit sequence by committing the other change out of order.  At the very least we need this re-based.

joehandzik:
@tasleson No, you didn't mess up the order, either one would have required a rebase because of the order I tacked on the capabilities.

I agree with you about the intended behavior point though, I asked about that in one of the other PRs where I brought this feature up. @cathay4t suggested that I might be able to leverage this existing API, but I agree with you that the original spirit of the API was an administrative enable/disable. That functionality doesn't exist for Smart Array though, it will always be a health-based enable/disable (and the disable is automatic). I'm fine either way, but I'd like to know sometime today or tomorrow so I have time to write a new API call if I need to...trying to get all the code I need finished before close of business tomorrow so we have all next week to get outstanding pulls in and fix bugs if any exist.

tasleson:
@joehandzik  From the code documentation `Makes a volume online and available to the host.` This is ambiguous enough that I think @cathay4t is correct in that we can use it for this functionality :-)

joehandzik:
@tasleson Then use it we shall. :) I just rebased, we can either wait for @cathay4t to chime in tonight or we can pull it in once the checks complete.

cathay4t:
As @joehandzik stated, after disk replaced, pool is in `unhealthy` state. Using `volume_enable()` to change pool state seems above its pay-grade.
I am thinking something like `pool_recheck()` to enforce pool do reconstruct(RAID 1/5/6/etc) or reenable (RAID 0).
How's that sounds?

joehandzik:
Well, at least for Smart Arrays, the rebuild happens automatically on any volume that can be degraded (for example, once a failed R1 volume is replaced, the data will be re-mirrored automatically). So the API wouldn't enforce anything, the reconstruction is standard.

volume_enable in its current state would pretty much just be used for volumes that the Smart Array firmware outright failed. The main use case would obviously be R0 disks, because if a RAID 1 or RAID 5 volume failed completely, some poor sysadmin is probably trying to actually recover data from the disks and wouldn't want to just keep moving along...

I'm fine either way. If you want a new API call, let me know if you have specific design constraints you'd like me to follow.

tasleson:
@joehandzik @cathay4t If this operation is typically automatic for smart arrays and is only done when the firmware fails, is that functionality needed in this API?  I would think in these rare situations one would be best advised to follow vendor documentation and utilize the vendor CLI directly to resolve?

joehandzik:
The operation I am proposing is not terribly rare in single disk RAID 0 AND environments, unfortunately. Having an API through libstoragemgmt would alleviate a lot of customer pain...I have specific requests from large customers to integrate this specific operation into scale-out solutions. I propose that we take this PR as-is, since smart array can't execute the administrator enable/disable operation that other arrays can.

Joe

On Apr 25, 2016, at 2:52 PM, Tony Asleson <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> @cathay4t<https://github.com/cathay4t> If this operation is typically automatic for smart arrays and is only done when the firmware fails, is that functionality needed in this API? I would think in these rare situations one would be best advised to follow vendor documentation and utilize the vendor CLI directly to resolve?

-

joehandzik:
@tasleson To clarify, I think I'm agreeing with you; the operation that @cathay4t describes is not something that makes a ton of sense for Smart Array. But I do want to make sure that everyone understand the use case I've described has merit beyond my own goals for functionality...we have lots of customers trying to use Smart Arrays in a manner more like how HBAs are used (no redundancy on the single volume/disk level, when something fails throw it away, replace it, and regenerate the data across the network).

tasleson:
@joehandzik @cathay4t I think for this specific use case, using the existing `volume_enable` call seems reasonable.

joehandzik:
@cathay4t Do you agree with us on this one? Once you give the ok, we can pull this feature in.

cathay4t:
+1 for `volume_enable()` on RAID 0 volume.
BTW, could you also update `lsm.Client.volumes()` to set `admin_state` as  `ADMIN_STATE_DISABLED` for broken RAID 0 volume so sure could know when to use `volume_enable()`. Thank you.

joehandzik:
@cathay4t Done. Also, this adequately serves the requirements that I had for volume status, so everyone wins!

tasleson:
Thanks @joehandzik

Issue #106, opened 2016-04-04
Request for comments: Option flags/settings for creating arrays and volumes


joehandzik:
Hey guys,

First, I have a new requirement. The same user that wants the read/write cache control also would need specific control over some options for creating volumes and arrays on Smart Array. For example, they create volumes on SSDs, but want to continue to use the controller BBWC with those SSDs (it just makes sense for their IO pattern). So, the command line option using hpssacli would look something like this for their requirement:

hpssacli ctrl slot=0 create type=ld drives=1i:1:13 size=max raid=0 ss=64 aa=enable

That "aa=enable" chunk is the only part missing in libstoragemgmt today. Now, I could go and create a command in libstoragemgmt called lsm_volume_raid_create_with_aa, but it feels to me like we could create a more extensible method...would the right answer be to add another variable to accept slight tweaks like this? Or, would we be able to leverage the existing "flags" variable to accomplish what I'm suggesting?

Now that I've written this, I guess I'm proposing that we use the existing "flags" variable, and create a new flag type every time a slight modification variant crops up for existing commands.

Sound ok? If so, I'll probably work on that this week, along with the feature to set read cache percentage. I'm also thinking that Gris may be better off leveraging this method for whatever issue he ran into while creating a volume on an SSD with Smart Array, rather than always using the "force" option.

tasleson:
@joehandzik The flags parameter is intended to handle small tweaks to the API, but it really shouldn't be utilized for vendor specific features.  Is `aa` aka. `array accelerator` really just marketing speak for `cache`?  If that's true then I would be OK with adding a generic parameter flag value(s) for enable/disable use of system cache during volume creation.  However, from a generic API standpoint wouldn't a user want to be able to configure this for pre-existing volume(s) too, to enable/disable?

What does the acronym `BBWC` represent?

joehandzik:
@tasleson Yup, aa is HPE-boy-aren't-we-special speak for the controller cache (or BBWC, if you will :D). It's relevant for SSD arrays and volumes in particular, because something else is enabled by default on those (a mode where IOs are sent directly to the SSD when possible). Since it happens by default, I don't really think we need a flag that forces the bypass mode to be enabled, but effectively I'd want a mode to ensure that controller caching happens for a given volume.

[...]

PR #124, opened 2016-04-22
libstoragemgmt flags: add flags to provide volume creation options

joehandzik:

This attempts to address #106. All names are negotiable, but let me know if the concept is acceptable or unacceptable. 

From my perspective, pros:

* keeps the code quite simple
* prevents adding APIs for slight tweaks to existing APIs
* remains extensible and can be applied to non-HPE tech

cons:

* some flags may not map to all systems. How should that be handled and/or messaged to the user?

As always, thoughts are appreciated. This is the last feature I need in 1.3, once this and all outstanding feature pulls go in I think we're in very good shape for all use cases I'm trying to address.

### DIFF
--- a/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
+++ b/c_binding/include/libstoragemgmt/libstoragemgmt_types.h
@@ -38,6 +38,16 @@ typedef uint64_t lsm_flag;
 #define LSM_CLIENT_FLAG_RSVD 0
 
 /**
+ * Options for volume settings at creation time.
+ * Bit field to support multiple settings for systems
+ * that can support them simultaneously.
+ */
+#define LSM_CLIENT_FLAG_VOLUME_CREATE_USE_SYSTEM_CACHE		0x00000001
+#define LSM_CLIENT_FLAG_VOLUME_CREATE_USE_IO_PASSTHROUGH	0x00000002
+#define LSM_CLIENT_FLAG_VOLUME_CREATE_DISABLE_SYSTEM_CACHE	0x00000004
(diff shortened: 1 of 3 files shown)


tasleson:
@joehandzik @cathay4t My only question is should the flag be an enumerated value type or a bitfield.  For example would you want to be able to enable two or more options at the same time or a combination of them?

joehandzik:
For this particular set of flags, you'd only use one or the other. But I don't know how other arrays work. There may be cases where you'd want to use multiple flags at once. Would you want separate enums for different logical combinations of flags, or a single enum for all flags? If it's the latter, I say we leave the code I just wrote as-is. If it's the former, that could work, but then we have multiple sets of flags...I don't know if that would be a problem in the future or not.

tasleson:
@joehandzik @cathay4t I'm leaning toward bitfield values, so you can set multiple conditions without requiring enumerate values to handle every possible combination.  However, bitfields aren't without API confusion.  This change set wouldn't need much change to work this way.  Either way I think we should have separate flag variables defined for each function.  Having a single enum for all would be too limiting IMHO, if I am understanding you correctly.

tasleson:
@veillard How is libvirt using the `flags` argument, is it an enumerated value or a bitfield?

joehandzik:
So, I just need to update the flags that I've defined to correspond to bit positions instead of integers to enable multiple flags, right? I'll be adding one more flag, based on an initial look for the customer that will be using this.

joehandzik:
@tasleson Updated as a bitfield without an enum wrapper. Note that my checks in the Smart Array plugin intentionally look for equivalence instead of checking bits because setting more than one of those flags shouldn't happen.

tasleson:
@joehandzik Looks good, thanks!

PR #127, opened 2016-04-27
smispy plugin: Numerous improvements

tasleson:
This PR addresses the following issues:

1. We were getting errors on one of the SMI-S system in the lab.  Added the case where CopyType ==  dmtf.ST_CONF_CAP_COPY_TYPE_UNSYNC_ASSOC we will then do a call to `ModifyReplicaSynchronization` to return to storage pool (19).  Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1323789
2. Capability was incorrect as we were comparing a function address to an integer
3. On volume delete we need to make sure volume is not masked before we try to delete it, so that we can return the correct error code instead of a plugin bug exception
4. Misc. pep8 clean-up, network error handling and code clean-up

### DIFF
--- a/plugin/smispy/utils.py
+++ b/plugin/smispy/utils.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2014 Red Hat, Inc.
+# Copyright (C) 2014-2016 Red Hat, Inc.
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
 # License as published by the Free Software Foundation; either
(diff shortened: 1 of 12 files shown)


plugin/smispy/utils.py
```diff
@@ -42,6 +42,13 @@ def cim_wrapper(*args, **kwargs):
                     if 'Errno 113' in desc:
                         raise LsmError(ErrorNumber.NETWORK_HOSTDOWN,
                                        'Host is down')
+                    if 'Errno 104' in desc:
+                        raise LsmError(ErrorNumber.NETWORK_CONNREFUSED,
+                                   'Connection reset by peer')
```
cathay4t: PEP8 error on this line:
```
utils.py:47:36: E128 continuation line under-indented for visual indent
```

tasleson: Yep, will update.

tasleson:
@cathay4t Please review

cathay4t:
@tasleson I will review and test them this weekend.

cathay4t:
Looks good and tested on EMC VNX. Merged. Thanks Tony!

PR #105, opened 2016-04-01
[PATCH V3 0/7]  Library API: Controlling IDENT/FAULT LED of local disk.

cathay4t:
* Only support SES(SCSI Enclosure Service) yet.
* New functions:
    * C:
        * `lsm_local_disk_ident_led_on()`
        * `lsm_local_disk_ident_led_off()`
        * `lsm_local_disk_fault_led_on()`
        * `lsm_local_disk_fault_led_off()`
    * Python:
        * `lsm.LocalDisk.ident_led_on()`
        * `lsm.LocalDisk.ident_led_off()`
        * `lsm.LocalDisk.fault_led_on()`
        * `lsm.LocalDisk.fault_led_off()`
* Tested on mpt2sas and hpsa enclosure.

Issue #79, opened 2015-12-15
Add depenency file


cathay4t:
Files user could use to install depenency:
```
# For RHEL/Centos/Fedora
sudo yum install `cat rh_rpm_depenency`
# For Debian/Ubuntu
sudo apt-get install `cat deb_depenency`
# For suse
sudo zypper in `cat suse_rpm_depenecy`.
```

joehandzik:
I'm voting we take care of this in 1.4. :)

cathay4t:
:+1:  for postponing to 1.4

cathay4t:
Change to 1.3 to make my document life easier.

cathay4t:
Closing as PR #137 fixed this.

joehandzik:
@cathay4t I was just trying to save work if we didn't have time to get it in, glad you had time!

PR #137, opened 2016-05-03
Misc fixes for 1.3

cathay4t:
**Based on PR #105, please review it first**

* Unify the optional property return values:
    ```
For numerical return, -2 for no support, -1 for unknown(bug, invalid argument).
For pointer return, NULL for no support, unknown, invalid argument.
    ``` 

* Eliminate the use of output pointer for property getter.

* Fix all compile warnings found by `csbuild` except:
 *  All warnings for `tools/udev/scan-scsi-target.c`. I am not quite familiar it yet.
 * All warnings for `daemon/lsm_rest.c`. It has been not maintained for a while. Maybe remove it?
 * All 'unused argumnet` warnings. They are hard to fix or won't fix.

* MegaRAID:
 * Add support of `volume_delete()`.
 * Use VPD83 as `lsm.Volume.id` if possible(when user is using new vpd83 supported storcli).

* Remove REST API daemon.

* Add dependency files. #79


cathay4t:
Might be more patch after I got result from `covscan`.

tasleson:
@cathay4t thanks Gris!

PR #139, opened 2016-05-05
Fixes for lsmcli disk_location and simarray disk_location values

### LINKED ISSUE #109: disk.location should be accessable by lsmcli.
BTW, all the disks in sim plugin are using the same location, try to change it meanwhile.

Just note down, I(Gris) will take care of this issue.

joehandzik:
Fixes #109


joehandzik:
I just noticed that pull #137 changes the definition of disk_location to location. I'll redo this pull to be in-line with that when #137 goes in.

tasleson:
@joehandzik Thanks!

PR #141, opened 2016-05-09
lsmcli and simarray: Fix disk_location usage to reflect change to loc…

joehandzik:
…ation

The name of the 'disk_location' attribute was changed to 'location', updating remaining uses of 'disk_location'.

### DIFF
--- a/plugin/sim/simarray.py
+++ b/plugin/sim/simarray.py
@@ -122,7 +122,7 @@ class PoolRAID(object):
 
 
 class BackStore(object):
-    VERSION = "3.8"
+    VERSION = "3.9"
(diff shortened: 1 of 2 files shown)


tasleson:
@joehandzik I think you should bump the db version for this too.

joehandzik:
@tasleson Done. In general, when should that VERSION value be changed? When elements of the database are added/deleted/changed?

tasleson:
@joehandzik I would say anytime a schema change occurs or a change that will cause the code unexpectedly to fail.

cathay4t:
Merged. Thanks Joe!

Issue #144, opened 2016-05-11
HPSA: Got 'Other/Not Configured' system status 


cathay4t:
Got a system with 'Cache Status: Not Configured' which cause `lsm.System.status == Other`.

```
Smart Array P440ar in Slot 0 (Embedded)
   Bus Interface: PCI
   Slot: 0
   Serial Number: <hidden>
   Cache Serial Number: <hidden>
   RAID 6 (ADG) Status: Enabled
   Controller Status: OK
   Hardware Revision: B
   Firmware Version: 3.00
   Rebuild Priority: High
   Expand Priority: Medium
   Surface Scan Delay: 3 secs
   Surface Scan Mode: Idle
   Parallel Surface Scan Supported: Yes
   Current Parallel Surface Scan Count: 1
   Max Parallel Surface Scan Count: 16
   Queue Depth: Automatic
   Monitor and Performance Delay: 60  min
   Elevator Sort: Enabled
   Degraded Performance Optimization: Disabled
   Inconsistency Repair Policy: Disabled
   Wait for Cache Room: Disabled
   Surface Analysis Inconsistency Notification: Disabled
   Post Prompt Timeout: 15 secs
   Cache Board Present: True
   Cache Status: Not Configured
   Cache Ratio: 100% Read / 0% Write
   Read Cache Size: 0 MB
   Write Cache Size: 0 MB
   Drive Write Cache: Disabled
   Total Cache Size: 2.0 GB
   Total Cache Memory Available: 1.8 GB
   No-Battery Write Cache: Disabled
   SSD Caching RAID5 WriteBack Enabled: True
   SSD Caching Version: 2
   Cache Backup Power Source: Batteries
   Battery/Capacitor Count: 1
   Battery/Capacitor Status: OK
   SATA NCQ Supported: True
   Spare Activation Mode: Activate on physical drive failure (default)
   Controller Temperature (C): 24
   Cache Module Temperature (C): 21
   Number of Ports: 2 Internal only
   Encryption: Disabled
   Express Local Encryption: False
   Driver Name: hpsa
   Driver Version: 3.4.10
   Driver Supports HPE SSD Smart Path: True
   PCI Address (Domain:Bus:Device.Function): 0000:03:00.0
   Negotiated PCIe Data Rate: PCIe 3.0 x8 (7880 MB/s)
   Controller Mode: RAID
   Controller Mode Reboot: Not Required
   Latency Scheduler Setting: Disabled
[...]

PR #145, opened 2016-05-11
HPE Smart Array: Return STATUS_OK when a cache module reports as unconfigured

joehandzik:

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -67,13 +67,14 @@ def _sys_status_of(hp_ctrl_status):
     check_list = [
         'Controller Status', 'Cache Status', 'Battery/Capacitor Status']
     for key_name in check_list:
-        if key_name in hp_ctrl_status and hp_ctrl_status[key_name] != 'OK':
-            # TODO(Gris Ge): Beg HP for possible values
-            status = System.STATUS_OTHER
+        if key_name in hp_ctrl_status:
+            if hp_ctrl_status[key_name] == 'Not Configured':
+                status = System.STATUS_OK
+            elif hp_ctrl_status[key_name] == 'OK':
+                status = System.STATUS_OK
+            else:
+                status = System.STATUS_OTHER
             status_info += hp_ctrl_status[key_name]
-
-    if status != System.STATUS_OTHER:
-        status = System.STATUS_OK
 
     return status, status_info
 


joehandzik:
Fixes #144

joehandzik:
One thing to note here is that this refactor is pulling a status_info now for the SYSTEM_OK status. Is that going to be a problem? If so, it can be fixed pretty easily by pulling the status_info population into the appropriate if branches.

cathay4t:
@joehandzik If 'not configured' only means no volumes are using cache in HPE SmartArray. How about we just treat it as SYSTEM_OK?

I got a system with only 2 ssd disks as RAID1 volume, with SmartPath enabled, RAM cache will be disabled. With this configure, I got 'no configure' cache status.
IMHO, this configure is OK and healthy, showing it as 'SYSTEM_UNCONFIGURED' sounds like user attention needed.

joehandzik:
@cathay4t I see your point, but calling it SYSTEM_OK loses some of the detail around the actual status. Maybe a compromise? I'll switch it to SYSTEM_OK for this config, but I'll continue to collect status_info for SYSTEM_OK statuses too. That way, a user could find out that the cache is actually unconfigured if they look in the right place.

joehandzik:
@cathay4t Compromise implemented here. Lemme know if you want it changed, otherwise this should cover your requirements and mine.

tasleson:
@joehandzik  @cathay4t System status is a bitfield, we can express that it's OK and that something else is true too.

tasleson:
@joehandzik Your latest commit comment doesn't match what the code change is.

joehandzik:
Yikes, you're very right. Will fix as soon as I can.

Joe

On May 11, 2016, at 10:41 AM, Tony Asleson <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> Your latest commit comment doesn't match what the code change is.

-

joehandzik:
@tasleson Ok, finally got the mismatched commit messaging addressed. 

It looks like we handle bitmaps correctly in the cli (from a quick glance, anyway), is there anywhere else we'd need to be concerned about? I'm fine with this not being in both an OK state and an unconfigured state for now, but we might want to think about it for 1.4.

tasleson:
@joehandzik The code should be handling bitfields, I don't know of any reasons why we should be concerned.

tasleson:
@joehandzik Thanks!

PR #146, opened 2016-05-11
test_automated.py: Send SIGKILL instead of SIGTERM

tasleson:
We are leaving processes hanging around for plugin tests that are not completing in time.


tasleson:
Tested out fine.

PR #148, opened 2016-05-12
MegaRAID plugin: Support disks with no enclosure.

cathay4t:
* With some MegaRAID/PERC(for example PERC H700), disks have no
   enclosure ID[1] which use '/cx/sall show all J' for disk querying.
   This patch try both '/c0/eall/sall show all J' and
   '/c0/sall show all J' for disk querying.

 * Also fixed the volume_raid_create() to support empty enclosure ID disks.

 * Tested on both enclosure disk and non-enclosure disk.

[1]: Please be informed, the enclosure here does not mean physical
     enclosure, I(Gris) have a MegaRAID card with 4 SATA disks directly
     attached to it, they all still have enclosure ID.
     I am guessing some old firmware or old MegaRAID/PERC card does not
     support enclosure ID. Anyway, we support both ways.

### DIFF
--- a/plugin/megaraid/megaraid.py
+++ b/plugin/megaraid/megaraid.py
@@ -525,13 +525,29 @@ class MegaRAID(IPlugin):
               flags=Client.FLAG_RSVD):
         rc_lsm_disks = []
         mega_disk_path_regex = re.compile(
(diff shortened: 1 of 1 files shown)


tasleson:
@cathay4t Thanks Gris!

PR #150, opened 2016-05-13
Fix RHEL/Fedora mock build test failure.

cathay4t:
Problem:
        When running in fakeroot, lsm_local_disk_list() might indicate the
        existence of '/dev/sda' while lsm_ident_led_on() fail with
        NOT_FOUND_DISK error.
    
    Root cause:
        The fakeroot might only have /sys mounted while /dev/ is not.
    
    Fix:
        Make sure the file '/dev/sdX' exists before add it to return list.

Also include a extra patch to remove unstable note on 1.3 plugin interface.


cathay4t:
Add another fix for RPM SPEC.

tasleson:
@cathay4t Thanks Gris!

PR #152, opened 2016-05-15
Misc fixes

cathay4t:
* Fix a bug introduced by PR #150.
 * Add new configure option to skip memory leak check test.
 * Fix lsm_local_disk_vpd83_search() for RHEL/Fedora mock build.

### DIFF
--- a/c_binding/utils.c
+++ b/c_binding/utils.c
@@ -68,14 +68,13 @@ void _be_raw_to_hex(uint8_t *raw, size_t len, char *out)
 
 bool _file_exists(const char *path)
 {
-    int fd = 0;
+    int fd = -1;
 
     assert(path != NULL);
 
     fd = open(path, O_RDONLY);
-    if (fd == -1) {
+    if ((fd == -1) && (errno == ENOENT))
         return false;
-    }
     close(fd);
     return true;
 }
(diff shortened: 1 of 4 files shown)


cathay4t:
@tasleson Sorry, we might need 1.3.2 release as I introduced a regression bug into 1.3.1.

cathay4t:
Found more issue on lsm_local_disk.c, fixing.

cathay4t:
New patch included for lsm_local_disk_vpd83_search() in RHEL/Fedora mock build.

c_binding/utils.c
```diff
@@ -68,14 +68,13 @@ void _be_raw_to_hex(uint8_t *raw, size_t len, char *out)
 
 bool _file_exists(const char *path)
 {
-    int fd = 0;
+    int fd = -1;
 
     assert(path != NULL);
 
     fd = open(path, O_RDONLY);
-    if (fd == -1) {
```
tasleson: I would suggest changing this to:

if ( fd == -1 && errno == ENOENT)

so that we are examining `errno` only when open returns an error, see:

http://www.gnu.org/software/libc/manual/html_node/Checking-for-Errors.html

for more information.

cathay4t:
Thank you for the note.

Fixed.

tasleson:
@cathay4t  Thanks Gris!

PR #154, opened 2016-05-18
Autotools: New configure option '--without-test'

cathay4t:
* Remove the old configure option '--disable-build-c-unit' as it will
   fail 'make check'.

 * Add new configure option './configure --without-test' to skip all
   test cases and their requirement check.

 * Add 'ps' into requirement as 'test_include.sh' needs it.


tasleson:
@cathay4t I added a commit to this branch, you cannot add whitespace in the if conditionals in the `Makefile.am` files.  I found this as `make check` never ran any test code when `--without-test` was not specified.

PR #156, opened 2016-06-15
ontap plugin: Correct error code for re-size

tasleson:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1346901


cathay4t:
According to [NetApp document][1], this problem is caused by limitation of NetApp `lun geometry`, Error `NO_SUPPORT` with error message `Manually change NetApp lun geometry is required for given size.` seems better.

[1]: https://kb.netapp.com/support/index?page=content&id=3012229&impressions=false

tasleson:
@cathay4t Yeah, that probably is more appropriate, I will redo change, thanks!

PR #157, opened 2016-06-16
ontap plugin: Correct error code for large re-size

tasleson:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1346901

More information:

https://kb.netapp.com/support/index?page=content&id=3012229


tasleson:
This change is incorrect, will redo again.

PR #159, opened 2016-06-20
Bug fix for sysfs vpd83 file not accessable.

cathay4t:
Please check https://bugzilla.redhat.com/show_bug.cgi?id=1346898 for detail.

This patch set is trying to use udev as fallback if sysfs vpd83 not available.


tasleson:
@cathay4t  Thanks Gris!

PR #160, opened 2016-06-20
Fix lsmcli job-status command on volume creation job.

cathay4t:
Problem:

    The 'lsmcli job-status --job <JOB_ID>' will fail on volume
    creation job due to missing 'sd_paths' property.

Fix:

    Call _add_sd_paths() before invoking display_data().


tasleson:
@cathay4t Thanks Gris!

PR #163, opened 2016-06-29
Bug fix: lsmcli volume-replicate-range will fail on C plugin.

cathay4t:
Problem:

    C plugin will raise/return INVALID_ARGUMENT error on this command:

        lsmcli volume-replicate-range --src-vol VOL_ID_00001 \
            --dst-vol VOL_ID_00002 --rep-type CLONE \
            --src-start 0 --dst-start 0 --count 100 -f

Root cause:

    lsmcli use string instead of int/long to initialize BlockRange
    object which cause value_to_block_range() in C library failed at
    range["src_block"].asUint64_t()

Fix:

    Convert string provided by argument parser to int.

### DIFF
--- a/tools/lsmcli/cmdline.py
+++ b/tools/lsmcli/cmdline.py
@@ -1488,7 +1488,8 @@ class CmdLine:
 
         ranges = []
         for b in range(len(src_starts)):
-            ranges.append(BlockRange(src_starts[b], dst_starts[b], counts[b]))
+            ranges.append(BlockRange(long(src_starts[b]), long(dst_starts[b]),
+                                     long(counts[b])))
 
         if self.confirm_prompt(False):
             self.c.volume_replicate_range(rep_type, src, dst, ranges)


cathay4t:
We need to find out a better fix in 1.4 release for:
 1. Raise better error message in C library when got illegal json data. 
     This kind of bug really takes quite a lot time to investigate.

 2. Should we enforce data type check in python `__init__()`?

This patch is just a quick fix.

cathay4t:
This problem is only reproducible using simulator C plugin from https://github.com/cathay4t/libstoragemgmt/tree/simc_sql branch

tools/lsmcli/cmdline.py
```diff
@@ -1488,7 +1488,8 @@ def volume_replicate_range(self, args):
 
         ranges = []
         for b in range(len(src_starts)):
-            ranges.append(BlockRange(src_starts[b], dst_starts[b], counts[b]))
+            ranges.append(BlockRange(int(src_starts[b]), int(dst_starts[b]),
+                                     int(counts[b])))
```
tasleson: C is treating this as an unsigned 64 bit integer.  When running python2 on 32bit hardware a python `int` is limited to 2**32-1.  We should be casting these to python `long`.

cathay4t:
Changed to long.

tasleson:
@cathay4t Thanks Gris!

Issue #97, opened 2016-03-05
Should not raise 'no deamon' error when checking help message of lsmcli.


cathay4t:
```
[fge@Gris-Laptop libstoragemgmt-code]$  lsmenv sim lsmcli  vc --help
Device alias: sim
URI: sim://
lsmcli vc --help
PLUGIN_IPC_FAIL(301): Unable to connect to lsmd, daemon started? 
Command exited with non-zero status 4
```

tasleson:
I looked at this one today and this one is interesting.  If you place the `--help` before the alias it works as expected.  For some currently unknown reason when `--help` is after the alias we return from `parse_known_args` with unknown args containing `--help`.  When we later try to call  `self.args.func(self.args)` we call into `handle_alias` which in turn calls `parse_args` again which catches that we have `--help` specified.  This of course only happens after we have established a connection with the daemon so we fail with the daemon present when really all we wanted was to display the help text.

Additionally, with this bug we are allowing random junk on the command line as long as it doesn't utilize an alias, eg. 

```
lsmcli list --type systems  --uri sim:// --we --don\'t --error --on --this --but --we --should
ID     | Name                          | Status | Info | FW Ver                                                  | Mode    | Read Cache Percentage
--------------------------------------------------------------------------------------------------------------------------------------------------
sim-01 | LSM simulated storage plug-in | OK     |      | LSM_SIMULATOR_DATA_3.9_47588d4b158a37889ddf44eeb8e88b0f | HW RAID | 10
```

I fear that we are abusing the argparse library excessively, to add features that it doesn't natively support which is causing us pain with issues like this one and when trying to migrate to py3.

PR #165, opened 2016-07-06
cmdline.py: Alternative way to handle aliases

tasleson:
@cathay4t Please take a look at this to see what you think.  I wasn't able to twiddle the argparse library to get things to work correctly, but maybe you will have better luck.  Maybe I'm missing something obvious, but this change is very straight forward and appears to restore things to a working state.  Address issue #97 

I fear we have out grown argparse with some of the things we are trying to do.  I know the py3 conversion has some ugly behavior with our use and abuse of argparse.

### DIFF
--- a/tools/lsmcli/cmdline.py
+++ b/tools/lsmcli/cmdline.py
@@ -867,44 +867,44 @@ cmds = (
     ),
 )
 
-aliases = (
-    ['ls', 'list --type systems'],
-    ['lp', 'list --type pools'],
-    ['lv', 'list --type volumes'],
-    ['ld', 'list --type disks'],
-    ['la', 'list --type access_groups'],
-    ['lf', 'list --type fs'],
-    ['lt', 'list --type target_ports'],
-    ['c', 'capabilities'],
-    ['p', 'plugin-info'],
-    ['vc', 'volume-create'],
-    ['vrc', 'volume-raid-create'],
-    ['vrcc', 'volume-raid-create-cap'],
-    ['vd', 'volume-delete'],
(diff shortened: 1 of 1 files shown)


cathay4t:
Looks good, but failed to handle connection free commands:

```bash
[fge@Gris-Laptop ~]$ sudo lsmenv sim lsmcli local-disk-list
Device alias: sim
URI: sim://
PLUGIN_IPC_FAIL(301): Unable to connect to lsmd, daemon started? 
Command exited with non-zero status 4
```

Patch:
```patch
diff --git a/tools/lsmcli/cmdline.py b/tools/lsmcli/cmdline.py
index 8258373..24f18c9 100644
--- a/tools/lsmcli/cmdline.py
+++ b/tools/lsmcli/cmdline.py
@@ -1689,11 +1689,8 @@ def is_connection_free_cmd(self):
         """
         Return True if current command is one of _CONNECTION_FREE_COMMANDS.
         """
-        if self.args.func == self.handle_alias and \
-           self.args.cmd[0] in _CONNECTION_FREE_COMMANDS:
-            return True
-
-        if self.args.func.__name__ in _CONNECTION_FREE_COMMANDS:
+        if self.args.func.__name__.replace("_", "-") in \
+           _CONNECTION_FREE_COMMANDS:
             return True
         return False

```

tasleson:
@cathay4t Nice catch.  Looks like we need to add a test case so we catch stuff like this with the unit test.

cathay4t:
Issue #166 created for tracking test case for connection free lsmcli commands.

PR #174, opened 2016-07-20
plugin/hpsa: Update possible binary locations for SSA tools

joehandzik:


joehandzik:
Quick update while I was poking around in the neighborhood. I found an unreleased version of hpssacli that dropped the "hp" prefix. I'm sure we'll end up with that release in the wild within the next six months or so...

cathay4t:
Merged. Thanks.

PR #177, opened 2016-08-01
hpsa: Update cacheless variable

joehandzik:
There was a mistake here that I ran into on an H240 Smart Array card in RAID mode. I used the wrong naming scheme, missed it because my hardware was not available until recently. 

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -469,7 +469,7 @@ class SmartArray(IPlugin):
             else:
                 # Some Smart Arrays don't have cache
                 # This entry is also missing until a volume uses cache
-                read_cache_pct = System.CACHE_PCT_UNKNOWN
+                read_cache_pct = System.READ_CACHE_PCT_UNKNOWN
             if 'Controller Mode' in ctrl_data:
                 hwraid_mode = ctrl_data['Controller Mode']
                 if hwraid_mode == 'RAID':


joehandzik:
Also, making a note of issues like this slipping in...this sort of error could be caught more easily if I could get you guys access to a dedicated HPE box somehow.

tasleson:
@joehandzik I'm working on changing the CI service components so that we can have multiple nodes which connect to a single coordinating node which is used as the contact point with github.  The existing code is: https://github.com/tasleson/lsm-ci .  The current architecture is too cumbersome as the end nodes require public connectivity (Restful API which uses ngrok).  Hopefully I will have something better by the end of the week that we can leverage at multiple labs so that for every pull request we can run our tests automatically and that is self healing and tolerates intermittent node failures better.

tasleson:
I'm going to double check the automated runs at the SNIA lab.  As it looks like we should re-role a release to pick up this change, is there anything else that needs a sanity test?  The LED's be blinking OK right? :-)

joehandzik:
@tasleson Still gotta check that...I'll do that now.

joehandzik:
@tasleson Confirmed, LEDs work just fine!

tasleson:
@joehandzik Thanks!  I'm going to roll a new release tomorrow AM, unless something else shows up when I'm looking through the logs at the lab.

cathay4t:
This kind of bug could be found by pylint. I will try to enable pylint check in CI #180.

For this patch, it looks good. Merged. Thanks.

PR #181, opened 2016-08-04
C library: Fix missing vpd83 optional property of disk.

cathay4t:
Problem:
    lsm_disk_vpd83_set() does works when transferring data to Python
    client API(like lsmcli).

Root cause:
    disk_to_value() does not handle vpd83.


cathay4t:
You need simc plugin from https://github.com/cathay4t/libstoragemgmt/tree/simc_sql to reproduce this.

joehandzik:
LGTM. Bummer that we missed this one.

PR #188, opened 2016-08-16
C library: Fix anon_uid/anon_gid from C client to plugin.

cathay4t:
Problem:

    When C client API passing LSM_NFS_EXPORT_ANON_UID_GID_NA(-1) to
    plugin, it will be converted to some large number[1].

Fix:

    Change lsm_mgmt.c to use -1 or -2 for JSON in IPC.

[1]: On RHEL 7, it will be UINT64_MAX - 1, on RHEL 6, it will be
     INT64_MAX - 1. For C++ code, I have no idea why the difference.


tasleson:
@cathay4t Thanks!

PR #190, opened 2016-08-16
Fix hpsa raid mode

### LINKED ISSUE #189: hpsa plugin bug: controllers in "RAID Mode" not supported
I have a Smart HBA H244br in one of my test systems that reports its mode as `RAID Mode` instead of the `RAID` option supported by libstoragemgmt currently. I'll go ahead and make a pull to fix this. Is there any way this fix can get incorporated in a release within the next few weeks?

**Offending output snippet**
```
Smart HBA H244br in Slot 0 (Embedded) (RAID Mode)
   Bus Interface: PCI
   Slot: 0
   Serial Number: PDNMC%%LM8W07W
   Cache Serial Number: PDNMC%%LM8W07W
   Controller Status: OK
   Hardware Revision: B
   Firmware Version: 4.02
   Rebuild Priority: High
   Expand Priority: Medium
   Surface Scan Delay: 3 secs
   Surface Scan Mode: Idle
   Parallel Surface Scan Supported: Yes
   Current Parallel Surface Scan Count: 1
   Max Parallel Surface Scan Count: 16
   Queue Depth: Automatic
   Monitor and Performance Delay: 60  min
   Elevator Sort: Enabled
   Degraded Performance Optimization: Disabled
   Inconsistency Repair Policy: Disabled
   Wait for Cache Room: Disabled
   Surface Analysis Inconsistency Notification: Disabled
   Post Prompt Timeout: 15 secs
   Cache Board Present: False
   Drive Write Cache: Disabled
   Controller Memory Size: 256 MB
   SATA NCQ Supported: True
   Spare Activation Mode: Activate on physical drive failure (default)
   Controller Temperature (C): 42
   Cache Module Temperature (C): 22
   Number of Ports: 1 Internal only
   Encryption: Disabled
   Express Local Encryption: False
   Driver Name: hpsa
   Driver Version: 3.4.14
   Driver Supports HPE SSD Smart Path: True
   PCI Address (Domain:Bus:Device.Function): 0000:04:00.0 
   Negotiated PCIe Data Rate: PCIe 3.0 x4 (3940 MB/s)
   Controller Mode: RAID Mode
   Controller Mode Reboot: Not Required
   Latency Scheduler Setting: Disabled
   Current Power Mode: MaxPerformance
   Sanitize Erase Supported: True
   Primary Boot Volume: None
   Secondary Boot Volume: None
```

BlaineEXE:
Fixes #189

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -477,9 +477,8 @@ class SmartArray(IPlugin):
                 elif 'HBA' in hwraid_mode:
                     mode = System.MODE_HBA
                 else:
-                    raise LsmError(
-                        ErrorNumber.PLUGIN_BUG,
-                        "Invalid Controller Mode: '%s'" % hwraid_mode)
+                    mode = System.MODE_UNKNOWN
+                    status_info += ' mode=[%s]' % str(hwraid_mode)
             else:
                 # prior to late Gen8, all Smart Arrays were RAID mode only
                 mode = System.MODE_HARDWARE_RAID
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -472,9 +472,9 @@ class SmartArray(IPlugin):
                 read_cache_pct = System.READ_CACHE_PCT_UNKNOWN
             if 'Controller Mode' in ctrl_data:
                 hwraid_mode = ctrl_data['Controller Mode']
-                if hwraid_mode == 'RAID':
+                if 'RAID' in hwraid_mode:
                     mode = System.MODE_HARDWARE_RAID
-                elif hwraid_mode == 'HBA':
+                elif 'HBA' in hwraid_mode:
                     mode = System.MODE_HBA
                 else:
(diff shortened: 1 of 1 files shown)


BlaineEXE:
The changes tested OK on my dev system.

cathay4t:
@BlaineEXE Thanks for the patch. When setting `System.MODE_UNKNOWN`, could you also append `hwraid_mode` string to `System.status_info` property? It might help user to report back when getting `System.MODE_UNKNOWN`.

tasleson:
Any issues with the system mode being unknown?

BlaineEXE:
I don't see any issues. I think supporting it is better than throwing the error. As a sort of contrived example, if our next gen controllers support USB keys and list a 'USB key' mode (it's a dumb example, I know), it would be much better from a forwards compatibility perspective to just say the mode is unknown since the enum already exists and then list the mode in the `status_info` property than it is to throw an error that might break a monitoring/management utility that ingests the info.

cathay4t:
Merged. Thanks.

PR #191, opened 2016-08-17
Improve readability of status_info from HPSA

BlaineEXE:
Since status_info may contain other information besides status, add
clarity to existing info. The form follows that implemented by the
system mode addition, and will now look like below example:

'status=['OK']'

If there is an unknown system mode, it will be changed as below example:

'status=['OK'] mode=['Unknown Mode']'

This will make it easy for other applications to parse and interpret
the outputs.


cathay4t:
Thanks. Merged.

Issue #202, opened 2016-09-06
How to handle capabilities that may not be supported in certain system modes


joehandzik:
HBA mode Smart Array controllers will fail the CI suite right now due to their inability to configure logical volumes (that's how our HBAs work right now, pure HBAs instead of the hybrid mode that some Avago controllers run in). There will likely be other examples of this, so...how should I handle pieces of the CI framework that will fail when a device is in a certain state, but are not indicative of failures? Is there a clean way to handle this, other than writing changes into the failing unit tests to check for the conditions that would make the failures acceptable? For example, I could check the system mode on failure of a volume create command, and if the system is in HBA mode then mark the test as passing. What I just suggested actually isn't that bad IMO, it's probably good for the tests to document when a failure might be justified...

tasleson:
@joehandzik The capabilities are suppose to be accurate, so that a user can tell if an operation is available.  I'm thinking that the plugin when the card is in HBA mode should not be returning that volume_create is available?

joehandzik:
@tasleson That's definitely an option. I suppose it depends on if you interpret the capabilities as being "capabilities of the plugin" or "capabilities of the device". Adjusting the cache module read/write ratio will be another one that requires some amount of upfront data...so, to make this work, I'll need to take a look at the system in self.capabilities, and make a decision on support based on whatever system information would inform that decision? Totally fine to handle it that way, just confirming that you agree.

tasleson:
@joehandzik The capabilities are suppose to be the capability of the device, sorry if we haven't documented this well enough to convey that.  For example SMI-S has a wide range of support across different arrays and we interrogate that at run-time and fill out the capabilities as appropriate.  Obviously the plugin needs to support the feature too, but ultimately the intent is what can I do with this hardware device right now.

joehandzik:
@tasleson It all makes sense and is totally reasonable. Hopefully PR #205 deals with the issue in an acceptable way.

PR #205, opened 2016-09-13
plugin/hpsa: Change capabilities based on system info

joehandzik:


joehandzik:
Fixes #202

joehandzik:
The change worked fine on my test system, the issue is that I wasn't running as root for the CI system initially...

tasleson:
@joehandzik Thanks!

Issue #203, opened 2016-09-07
Feature request - Serial Number for physical disks via local_disk API


joehandzik:
We have a partner that would like to extract disk serial numbers without using a vendor-specific plugin. This shouldn't be too tough, I'll get to it as soon as I can.

cathay4t:
The SCSI SPC `READ MEDIA SERIAL NUMBER` command should do. I will take care this.

cathay4t:
Opps. As sg3_utils indicates ` It is not an mandatory command and the author has not seen any SCSI devices that support it.`. Will find another way.

cathay4t:
vpd 0x80 `Unit Serial Number VPD page` could be used.

joehandzik:
That was the path I had started down (vpd80), yeah. It would look really similar you the existing vpd83 code. I have this half started and can probably finish it tomorrow...maybe I take this one and you poke at the LED states? If you already have code in progress too, it's alright if you beat me to this.

On Sep 7, 2016, at 10:07 PM, Gris Ge <notifications@github.com<mailto:notifications@github.com>> wrote:


vpd 0x80 Unit Serial Number VPD page could be used.

-

cathay4t:
No, I haven't code anything yet. You may continue on this. I will work on the SES LED feature.

joehandzik:
@cathay4t Sounds good!

PR #207, opened 2016-09-15
local_disk: Extract serial number from vpd80

joehandzik:
Fixes issue #203

Signed-off-by: Joe Handzik joseph.t.handzik@hpe.com

### DIFF
--- a/c_binding/libsg.c
+++ b/c_binding/libsg.c
@@ -114,6 +114,13 @@ struct _sg_t10_vpd83_header {
     uint16_t page_len_be;
 };
 
+struct _sg_t10_vpd80_header {
+    uint8_t dev_type : 5;
+    uint8_t qualifier : 3;
+    uint8_t page_code;
+    uint16_t page_len_be;
+};
+
 struct _sg_t10_vpd00 {
     uint8_t ignore;
     uint8_t page_code;
@@ -329,6 +336,62 @@ bool _sg_is_vpd_page_supported(uint8_t *vpd_0_data, uint8_t page_code)
             return true;
     }
     return false;
+}
+
+int _sg_parse_vpd_80(char *err_msg, uint8_t *vpd_data, uint8_t *serial_num,
+                     uint16_t serial_num_max_len)
+{
+    int rc = LSM_ERR_OK;
+    int len = 0;
+    struct _sg_t10_vpd80_header *vpd80_header = NULL;
+    uint8_t *p = NULL;
+    uint8_t *end_p = NULL;
+    uint16_t vpd80_len = 0;
+    uint16_t serial_num_len = 0;
+
+    assert(err_msg != NULL);
+    assert(vpd_data != NULL);
+    assert(serial_num != NULL);
+    assert(serial_num_max_len != 0);
+
+    memset(serial_num, 0, serial_num_max_len);
+
+    vpd80_header = (struct _sg_t10_vpd80_header*) vpd_data;
+
+    if (vpd80_header->page_code != _SG_T10_SPC_VPD_UNIT_SN) {
+        rc = LSM_ERR_LIB_BUG;
+        _lsm_err_msg_set(err_msg, "BUG: Got incorrect VPD page code '%02x', "
+                         "should be 0x80", vpd80_header->page_code);
+        goto out;
+    }
+
+    serial_num_len = be16toh(vpd80_header->page_len_be);
+    vpd80_len = serial_num_len + sizeof(struct _sg_t10_vpd83_header);
+
+    end_p = vpd_data + vpd80_len - 1;
+
+    if (end_p >= vpd_data + _SG_T10_SPC_VPD_MAX_LEN) {
+        rc = LSM_ERR_LIB_BUG;
+        _lsm_err_msg_set(err_msg, "BUG: Got invalid VPD UNIT SN page response, "
+                         "data length exceeded the maximum size of a legal VPD "
+                         "page");
+        goto out;
+    }
+
+    p = vpd_data + sizeof(struct _sg_t10_vpd83_header);
+
+    //add extra character to allow for terminating NULL
+    serial_num_len += 1;
+    len = snprintf((char *) serial_num, serial_num_len, "%s", (char *) p);
+
+    if ((uint16_t) len >= serial_num_len) {
+        memset(serial_num, 0, serial_num_max_len);
+        rc = LSM_ERR_LIB_BUG;
+        _lsm_err_msg_set(err_msg, "BUG: VPD UNIT SN was truncated when copied.");
+    }
+
+ out:
+    return rc;
 }
 
 int _sg_parse_vpd_83(char *err_msg, uint8_t *vpd_data,
--- a/c_binding/lsm_local_disk.c
+++ b/c_binding/lsm_local_disk.c
@@ -41,6 +41,8 @@
 #include "libsg.h"
 #include "libses.h"
 
+#define _LSM_MAX_SERIAL_NUM_LEN			        253
+/* ^ Max is 252 bytes */
 #define _LSM_MAX_VPD83_ID_LEN                   33
 /* ^ Max one is 6h IEEE Registered Extended ID which it 32 bits hex string. */
 #define _SYS_BLOCK_PATH "/sys/block"
@@ -52,6 +54,9 @@
 
 #define _SD_PATH_FORMAT "/dev/%s"
 #define _MAX_SD_PATH_STR_LEN 128 + _MAX_SD_NAME_STR_LEN
+
+#define _SYSFS_VPD80_PATH_FORMAT "/sys/block/%s/device/vpd_pg80"
+#define _MAX_SYSFS_VPD80_PATH_STR_LEN  128 + _MAX_SD_NAME_STR_LEN
 
 #define _SYSFS_VPD83_PATH_FORMAT "/sys/block/%s/device/vpd_pg83"
 #define _MAX_SYSFS_VPD83_PATH_STR_LEN  128 + _MAX_SD_NAME_STR_LEN
@@ -75,6 +80,11 @@ struct t10_sbc_vpd_bdc {
 
 #pragma pack(pop)
 
+static int _sysfs_serial_num_of_sd_name(char *err_msg,
+                                        const char *sd_name,
+                                        uint8_t *serial_num);
+static int _sysfs_vpd_pg80_data_get(char *err_msg, const char *sd_name,
+                                    uint8_t *vpd_data, ssize_t *read_size);
 static int _sysfs_vpd83_naa_of_sd_name(char *err_msg, const char *sd_name,
                                        char *vpd83);
 static int _udev_vpd83_of_sd_name(char *err_msg, const char *sd_name,
@@ -94,6 +104,55 @@ static int _sysfs_sas_addr_get(const char *blk_name, char *tp_sas_addr);
 
 static int _ses_ctrl(const char *disk_path, lsm_error **lsm_err,
                      int action, int action_type);
+
+/*
+ * Retrieve the content of /sys/block/sda/device/vpd_pg80 file.
+ * No argument checker here, assume all non-NULL and vpd_data is
+                * char[_SG_T10_SPC_VPD_MAX_LEN]
+ */
+static int _sysfs_vpd_pg80_data_get(char *err_msg, const char *sd_name,
+                                    uint8_t *vpd_data, ssize_t *read_size)
+{
+    int file_rc = 0;
+    char sysfs_path[_MAX_SYSFS_VPD80_PATH_STR_LEN];
+    char sysfs_blk_path[_MAX_SYSFS_BLK_PATH_STR_LEN];
+    char strerr_buff[_LSM_ERR_MSG_LEN];
+
+    memset(vpd_data, 0, _SG_T10_SPC_VPD_MAX_LEN);
+
+    /*
+     * Check the existence of disk vis /sys/block/sdX folder.
+     */
+    snprintf(sysfs_blk_path, _MAX_SYSFS_BLK_PATH_STR_LEN,
+             _SYSFS_BLK_PATH_FORMAT, sd_name);
+    if (! _file_exists(sysfs_blk_path)) {
+        _lsm_err_msg_set(err_msg, "Disk %s not found", sd_name);
+        return LSM_ERR_NOT_FOUND_DISK;
+    }
+
+    snprintf(sysfs_path, _MAX_SYSFS_VPD80_PATH_STR_LEN,
+             _SYSFS_VPD80_PATH_FORMAT, sd_name);
+
+    file_rc = _read_file(sysfs_path, vpd_data, read_size,
+                         _SG_T10_SPC_VPD_MAX_LEN);
+    if (file_rc != 0) {
+        if (file_rc == ENOENT) {
+            _lsm_err_msg_set(err_msg, "File '%s' not exist", sysfs_path);
+            return LSM_ERR_NO_SUPPORT;
+        } else if (file_rc == EINVAL) {
+            _lsm_err_msg_set(err_msg, "Read error on File '%s': "
+                             "invalid argument", sysfs_path);
+            return LSM_ERR_NO_SUPPORT;
+        } else {
+            _lsm_err_msg_set(err_msg, "BUG: Unknown error %d(%s) from "
+                             "_read_file().", file_rc,
+                             strerror_r(file_rc, strerr_buff,
+                                        _LSM_ERR_MSG_LEN));
+            return LSM_ERR_LIB_BUG;
+        }
+    }
+    return LSM_ERR_OK;
+}
 
 /*
  * Retrieve the content of /sys/block/sda/device/vpd_pg83 file.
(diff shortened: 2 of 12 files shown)


cathay4t:
@joehandzik Since this adds new interface, please bump version to 1.4 in `configure.ac` and function documents.

joehandzik:
Ok, configure.ac and all files where vpd80_get was mentioned as 1.3 feature have been updated to 1.4.

Worth mentioning that I did not invest the time into finding a workaround for environments that may not have the vpd80 sysfs attribute pulled already. My intent is to go through and actually support a true VPD80 inquiry at some point.

cathay4t:
@joehandzik Sorry for late response, and I am afraid this has to wait one more week(I am taking PTO next week Oct 1-7). Could you rebase to current master branch when you got a chance? Thank you.

joehandzik:
No worries @cathay4t, it's not something I'm in a huge rush on at the moment. I'll have some SMART Status work by the time you come back for you to look at too. I'll rebase the next time I pull the branch down.

c_binding/libsg.c
```diff
@@ -114,6 +114,15 @@ struct _sg_t10_vpd83_header {
...
 
+struct _sg_t10_vpd80_header {
+    uint8_t data_msb : 5;
+    uint8_t naa_type : 3;
+    uint8_t page_code;
+    uint8_t rsvd;
+    uint8_t page_len_be;
+    uint8_t serial_num[_SG_T10_SPC_SN_MAX_LEN];
+};
+
```
cathay4t: The `struct _sg_t10_vpd80_header` here is incorrect. According to SCSI SPC-4 rev37 and SPC-5 rev7, it should be:
```
struct _sg_t10_vpd80_header {
    uint8_t dev_type : 5;
    uint8_t qualifier : 3;
    uint8_t page_code;
    uint16_t page_len_be;
};
```
And you should not include `uint8_t serial_num[_SG_T10_SPC_SN_MAX_LEN]` in this struct as `serial_num` is not fixed length.

c_binding/lsm_local_disk.c
```diff
@@ -379,6 +483,67 @@ int lsm_local_disk_vpd83_search(const char *vpd83,
...
+
+    rc = _check_null_ptr(err_msg, 3 /* arg_count */, disk_path, vpd80, lsm_err);
+
+    if (rc != LSM_ERR_OK) {
+        if (vpd80 != NULL)
+            *vpd80 = NULL;
+
+        goto out;
+    }
+
```
cathay4t: Line 496-503 could be simplified to one line:
```
_good(_check_null_ptr(err_msg, 3 /* arg_count */, disk_path, vpd80, lsm_err), rc, out);
```

cathay4t:
Meanwhile, please also allow `lsmcli ldl -s` to query disk serial number(only show it when `-s`).

joehandzik:
Ok, adjustments made:

* removed serial number from vpd80 struct
* simplified the code section with _good() as suggested
* check for ASCII spaces as an initial character in the hex string pulled from VPD 80. Assuming that spaces are never valid characters for a serial number, a leading space should indicate that the entire string is ASCII spaces
* remove any trailing ASCII spaces by inserting a null terminator in place of the first trailing space, effectively preventing strdup from including all trailing spaces

The one thing I didn't do was add the 'ldl -s' option. Are you sure that's a good idea? I feel like it conflicts with the 'script' flag that already exists, and there's no precedent (that I could see, anyway) for requiring flags to get more data out of the 'list' commands. What value is there in preventing the serial number from being present in a basic 'lsmcli ldl' call?

joehandzik:
Had to rollback the _good() change, it's at least partially responsible for the valgrind errors that just showed up. See here for how this was done in the vpd83 code: https://github.com/libstorage/libstoragemgmt/blob/master/c_binding/lsm_local_disk.c#L394

cathay4t:
We move some properties to `-s` option due to console text width limitation.
For precedent, check `df -a` vs `df`

c_binding/lsm_local_disk.c
```diff
@@ -379,6 +483,72 @@ int lsm_local_disk_vpd83_search(const char *vpd83,
...
+    sd_name = disk_path + strlen("/dev/");
+
+    rc = _sysfs_vpd80_serial_num_of_sd_name(err_msg, sd_name, tmp_vpd80);
+    if (rc != LSM_ERR_OK)
+        goto out;
+
+    if (tmp_vpd80[0] != '\0' || tmp_vpd80[0] != 0x20) {
```
cathay4t: In stead of magic number `0x20`, please use `' '` instead.

c_binding/utils.c
```diff
@@ -114,3 +114,19 @@ int _read_file(const char *path, uint8_t *buff, ssize_t *size, size_t max_size)
     }
     return rc;
 }
+
+/* Find the first trailing space character,
+ * insert a NULL character in its place.
+ */
+void _trim_trailing_spaces(char *string_ptr)
+{
```
cathay4t: If `string_prt` is not terminated by `\0`[1], we may get a segfault here. Please consider to include a string length here.

[1]: I may happen on some buggy kernel driver or firmware, and SCSI standard said it **may** contain one or more `\0` as terminator

c_binding/utils.c
```diff
@@ -114,3 +114,19 @@ int _read_file(const char *path, uint8_t *buff, ssize_t *size, size_t max_size)
...
+    int i = 0;
+
+    while (string_ptr[i] != '\0') {
+        if (string_ptr[i] == 0x20) {
+            string_ptr[i] = '\0';
+            break;
+        }
+        ++i;
+    }
```
cathay4t: The SPC 3,4,5 indicates:

```
The PRODUCT SERIAL NUMBER field contains right-aligned ASCII data (see 4.3.1) that is a vendor 
specific serial number.
```

So You should remove both leading space[1] and trailing space. And meanwhile, the SPC allows space as serial numbe: 
```
ASCII data fields shall contain only ASCII printable characters (i.e., code values 20h 
through 7Eh) and may be terminated with one or more ASCII null (00h) characters.
```

[1]: The LSI expander use leading space padding:
```
[fge@megadeth ~]$ sudo sg_vpd --page=0x80 /dev/sde -H
Unit serial number VPD page:
 00     00 80 00 14 20 20 20 20  20 57 44 2d 57 43 43 34    ....     WD-WCC4
 10     4a 45 44 41 58 45 53 5a                             JEDAXESZ
[fge@megadeth ~]$ sudo sg_vpd --page=0x80 /dev/sde 
Unit serial number VPD page:
  Unit serial number:      WD-WCC4JEDAXESZ
```

cathay4t:
For the `_good()` usage, you are correct. I didn't notice the `free(*vpd80)` line.

joehandzik:
Ok, here's where I'm at now, @cathay4t:

* Understood regarding moving pieces under -s due to width constraints. I just haven't seen us do it in libstoragemgmt. df is a different tool, I was asking for a pointer to where we've done it in this source base (I didn't see an obvious example). If you still want me to try to move the serial number output out of the base command, I will, though I would argue that the serial number is more useful as a base attribute than, say, link rate or SAS/SATA.
* In my latest revision to this PR, I have removed the spaces function that I created entirely, along with any attempt to trim whitespace at all. Given what you just said about a space being a valid character in a serial number, I can't imagine that I should be attempting to trim out any space characters at all. For example, "  AABBCC" may be a fully valid serial number, and if I parse off those leading spaces I could be removing part of the actual serial number. So I think the safest move would be for us to leave all spaces alone. I can see some possible value in identifying cases where the entire serial number is spaces (you mentioned previously that it's a possible return value when the serial number is unavailable), so I could go back and add functionality to identify a string entirely composed of spaces. 

Any thoughts? Other than those two issues I think we're on the same page.

cathay4t:
@joehandzik 
* Example of hiding properties: `POOL_COLUMN_SKIP_KEYS` in `data_display.py`. If you think vpd80 is more valuable, please add a patch to hide less valuable information. @tasleson and I agreed on assuming lsmcli user console text width is 120(a recommendation, not mandatory).
* For trimming leading and trailing space, that's what udev/systemd did for `ID_SERIAL`. You may use functions mentioned in http://stackoverflow.com/questions/122616/how-do-i-trim-leading-trailing-whitespace-in-a-standard-way [which is MIT license][1].

[1]: https://meta.stackexchange.com/questions/272956/a-new-code-license-the-mit-this-time-with-attribution-required?cb=1

joehandzik:
@cathay4t 

1. Ok, I have a second commit in this branch to move disk RPM to a hidden attribute. Was pretty easy now that I know what the those *_SKIP_KEYS arrays are for...I'm open to discussing the merits of hiding something else, even serial number if you have a good argument for it.
2. There's a new attempt at trimming spaces in utils.c. Lemme know what you think.

c_binding/utils.c
```diff
@@ -114,3 +114,31 @@ int _read_file(const char *path, uint8_t *buff, ssize_t *size, size_t max_size)
     }
     return rc;
 }
+
+char* _trim_spaces(char *beginning)
+{
+    size_t len = 0;
+    unsigned int leading_space_count = 0;
+    char *end;
+
```
cathay4t: Please add `assert(beginning != NULL)`.

c_binding/utils.h
```diff
@@ -92,4 +92,11 @@ LSM_DLL_LOCAL bool _file_exists(const char *path);
 LSM_DLL_LOCAL int _read_file(const char *path, uint8_t *buff, ssize_t *size,
                              size_t max_size);
 
+/*
+ * Preconditions:
+ * beginning != NULL
+ *
+ * Return NULL if failed.
+ */
+LSM_DLL_LOCAL char* _trim_characters(char *beginning);
```
cathay4t: Should be `_trim_spaces()` here.

c_binding/lsm_local_disk.c
```diff
@@ -379,6 +483,80 @@ int lsm_local_disk_vpd83_search(const char *vpd83,
...
+    }
+
+    sd_name = disk_path + strlen("/dev/");
+
+    rc = _sysfs_vpd80_serial_num_of_sd_name(err_msg, sd_name, tmp_vpd80);
+    if (rc != LSM_ERR_OK)
+        goto out;
+
+    if (tmp_vpd80[0] != '\0') {
+        trimmed_vpd80 = _trim_spaces(tmp_vpd80);
```
cathay4t: Since you don't pass string length to `_trim_spaces()`, please add `tmp_vpd80[_LSM_MAX_VPD80_ID_LEN - 1] = '\0';` before invoke `_trim_spaces()` to make sure passing string is `\0` terminated.

c_binding/libsg.c
```diff
@@ -331,6 +338,55 @@ bool _sg_is_vpd_page_supported(uint8_t *vpd_0_data, uint8_t page_code)
...
+        _lsm_err_msg_set(err_msg, "BUG: Got invalid VPD UNIT SN page response, "
+                         "data length exceeded the maximum size of a legal VPD "
+                         "page");
+        goto out;
+    }
+
+    p = vpd_data + sizeof(struct _sg_t10_vpd83_header);
+
+    len = snprintf(serial_num, serial_num_len, "%s", p);
+
```
cathay4t: Please remove trailing whitespace here.

cathay4t:
@joehandzik I tried your patch on my laptop, but got `LIB_BUG` error:
```
[fge@Gris-Laptop libstoragemgmt-code]$ sudo lsmenv sim lsmcli ldl
Device alias: sim
URI: sim://
lsmcli ldl
LIB_BUG(1): BUG: VPD UNIT SN exceeded the preallocated size from caller. 
Command exited with non-zero status 4

Time: 0:00.09

Return code: 4
[fge@Gris-Laptop libstoragemgmt-code]$ hexdump  /sys/block/sda/device/vpd_pg80
0000000 8000 1400 3253 3958 584e 4830 3037 3436
0000010 3535 2020 2020 2020                    
0000018
```

joehandzik:
@cathay4t This should work now, FYI. I reproduced your issue on my side, looks like I hadn't successfully run against a device that was returning data via VPD 80 after the previous wave of fixes.

cathay4t:
The `lsmcli ldl` hang on my server with 100% cpu load. Not sure related or not. Will do code review and update founds tomorrow.

joehandzik:
@cathay4t It's strange that you keep running into issues on your laptop (not that they aren't legitimate, just strange). I've run several times against a couple of different DL380s without any observable issue. I'll take a look at this again when I have a chance tomorrow.

c_binding/lsm_local_disk.c
```diff
@@ -379,6 +483,80 @@ int lsm_local_disk_vpd83_search(const char *vpd83,
     return rc;
 }
 
+int lsm_local_disk_vpd80_get(const char *disk_path, char **vpd80,
+                             lsm_error **lsm_err)
+{
+    char tmp_vpd80[_LSM_MAX_VPD80_ID_LEN];
+    char *trimmed_vpd80;
```
cathay4t: Please initialize `char *trimmed_vpd80` to NULL.

c_binding/utils.c
```diff
@@ -114,3 +114,33 @@ int _read_file(const char *path, uint8_t *buff, ssize_t *size, size_t max_size)
...
+    assert(beginning != NULL);
+
+    len = strlen(beginning);
+
+    if (!len)
+        return NULL;
+
+    while (*beginning == ' ')
+        leading_space_count++;
```
cathay4t: This is where the infinite loop started.

c_binding/lsm_local_disk.c
```diff
@@ -220,6 +278,52 @@ static int _sysfs_vpd83_naa_of_sd_name(char *err_msg, const char *sd_name,
...
+static int _sysfs_vpd80_serial_num_of_sd_name(char *err_msg,
+                                              const char *sd_name, char *vpd80)
+{
+    ssize_t read_size = 0;
+    struct _sg_t10_vpd80_header *sn_header = NULL;
+    int rc = LSM_ERR_OK;
+    uint8_t vpd_data[_SG_T10_SPC_VPD_MAX_LEN];
+    uint16_t i = 0;
+
+    memset(vpd80, 0, _LSM_MAX_VPD80_ID_LEN);
```
cathay4t: Please move `memset()` into `_sg_parse_vpd_80()`.

c_binding/libsg.c
```diff
@@ -331,6 +338,54 @@ bool _sg_is_vpd_page_supported(uint8_t *vpd_0_data, uint8_t page_code)
     return false;
 }
 
+int _sg_parse_vpd_80(char *err_msg, uint8_t *vpd_data, uint8_t *serial_num)
```
cathay4t: Here you use `serial_num`, but caller use `vpd80`, can we be consistent for this?

cathay4t:
@joehandzik If you are follow the SCSI T10 standard correctly, there should be no problem if code them in right way. And the problem is not on my laptop, its on a LSI enclosure disk which is quit common in server market, they use leading space for vpd80, this is the hexdump:

```
[fge@megadeth ~]$ hexdump /sys/block/sdg/device/vpd_pg80
0000000 8000 1400 2020 2020 5720 2d44 4357 3443
0000010 4c4a 4659 4534 3959                    
0000018
```

Code review change requests:
 * Please consider to rename API to `lsm_local_disk_serial_get()`. The `vpd83` api is just to align with existing `lsm.Disk` class. For this new API, we might consider use non-tech name for it.
 * Please DO fix issue I mentioned in https://github.com/libstorage/libstoragemgmt/pull/207#discussion_r85052033
 * Fix the infinite loop in https://github.com/libstorage/libstoragemgmt/pull/207/commits/49d7119591e882e89db7cb583ce4987e756a36ba#r87527666 
 * Fix other trivial issues I mentioned in code review.

joehandzik:
@cathay4t Fair enough, apologies for assuming you were using your laptop. Any system that this code executes against is a totally valid use case, I think we agree here.

I believe that I have fixed every item that you mentioned (including the one I had missed from the previous round of reviews). The main issue I see here is that I don't have an easy path to unit test the space trimming code. We may need to create a broader strategy around unit testing to catch corner cases that would trigger code like the space trimming.

joehandzik:
@cathay4t Looks like you're running the megaraid CI code as non-root, that's why the megaraid node shows as failed.

cathay4t:
@joehandzik Thanks for the update. I will take a look soon.
The MegaRAID CI is running as root, but the failure indicates otherwise, I will look into it.

c_binding/lsm_local_disk.c
```diff
@@ -220,6 +279,53 @@ static int _sysfs_vpd83_naa_of_sd_name(char *err_msg, const char *sd_name,
...
+        _lsm_err_msg_set(err_msg, "_sysfs_serial_num_of_sd_name(): "
+                         "Input sd_name argument is NULL");
+        rc = LSM_ERR_LIB_BUG;
+        goto out;
+    }
+
+    _good(_sysfs_vpd_pg80_data_get(err_msg, sd_name, vpd_data, &read_size),
+          rc, out);
+
+    char tmp_serial_num[_LSM_MAX_SERIAL_NUM_LEN];
```
cathay4t: `char tmp_serial_num` is not used.

cathay4t: `sn_header` is unused variable.

c_binding/libsg.c
```diff
@@ -331,6 +338,59 @@ bool _sg_is_vpd_page_supported(uint8_t *vpd_0_data, uint8_t page_code)
...
+        rc = LSM_ERR_LIB_BUG;
+        _lsm_err_msg_set(err_msg, "BUG: Got invalid VPD UNIT SN page response, "
+                         "data length exceeded the maximum size of a legal VPD "
+                         "page");
+        goto out;
+    }
+
+    p = vpd_data + sizeof(struct _sg_t10_vpd83_header);
+
+    len = snprintf(serial_num, serial_num_len, "%s", p);
```
cathay4t: Compile warning, should be:

`len = snprintf(serial_num, serial_num_len, "%s", (char *) p);`

c_binding/lsm_local_disk.c
```diff
@@ -220,6 +279,53 @@ static int _sysfs_vpd83_naa_of_sd_name(char *err_msg, const char *sd_name,
...
+ * Input *serial_num should be char[_LSM_MAX_SERIAL_NUM_LEN], assuming caller
+ * did the check.
+ * The maximum *sd_name strlen is (_MAX_SD_NAME_STR_LEN - 1), assuming caller
+ * did the check.
+ * Return LSM_ERR_NO_MEMORY or LSM_ERR_NO_SUPPORT or LSM_ERR_LIB_BUG or
+ *        LSM_ERR_NOT_FOUND_DISK
+ */
+static int _sysfs_serial_num_of_sd_name(char *err_msg,
+                                              const char *sd_name,
+                                              char *serial_num)
```
cathay4t: Code is not well aligned.

cathay4t: The `char *serial_num` should be `uint8_t *serial_num`.

cathay4t:
Test passed[1] on my laptop and the server I got.

Just fix some trivial issues, then we are ready for merge.


[1]: Found a issue(`lsmcli ldl` per-mutural break from try-except.), but not this PR related. I will fix it in another PR.

joehandzik:
@cathay4t Thanks for the check, the little tidbits have been fixed (not sure why I don't see those warnings on my side, they're real issues). Feel free to pull in when you feel comfortable doing so.

c_binding/lsm_local_disk.c
```diff
@@ -379,6 +482,84 @@ int lsm_local_disk_vpd83_search(const char *vpd83,
     return rc;
 }
 
+int lsm_local_disk_serial_num_get(const char *disk_path, char **serial_num,
+                                  lsm_error **lsm_err)
+{
+    char tmp_serial_num[_LSM_MAX_SERIAL_NUM_LEN];
```
cathay4t: `char tmp_serial_num[_LSM_MAX_SERIAL_NUM_LEN];` should be
 `uint8_t tmp_serial_num[_LSM_MAX_SERIAL_NUM_LEN];`.

cathay4t:
@joehandzik I fixed the MegaRAID CI by commit https://github.com/libstorage/libstoragemgmt/commit/e7427ff986c91f1260e7129dd1781b6e1d5224b8 . And we are two trivial fixes away from merge.

joehandzik:
@cathay4t Two changes complete, so this should be good to go. Pull in when you get a chance tomorrow if you agree. Glad the MegaRAID bit was an easy enough fix. 

Also, maybe worth spinning off as a separate issue, but would it be reasonable to change the configure.ac file to ensure that the '-Wall' flag is enabled for gcc? I'd rather catch these warnings myself via the 'make' process than rely on you to be catching them. If not '-Wall', then whatever flags make sense.

joehandzik:
@cathay4t Do not merge yet, just found an issue via a functional test against a SAS drive. I'll update the PR with a fix when I have one and leave a message that we're good to merge again.

joehandzik:
@cathay4t Ok, the fix is in. I was accidentally truncating SAS serial numbers. I now allocate an extra byte of space before copying from the vpd80 page into a serial number buffer.

cathay4t:
@joehandzik 
For gcc `-Wall` option, I am enabling them in https://github.com/cathay4t/libstoragemgmt/tree/misc_fixes , but with that, some warning are treated as compile error, I am actively fixing those warnings. Will let you know once done.

For this PR, it still have a trival issue:
`libstoragemgmt-1.4.0/c_binding/lsm_local_disk.c:532:9: warning: pointer targets in passing argument 1 of '_trim_spaces' differ in signedness [-Wpointer-sign]`

But I am merging this PR and this issue will be fixed when I enable gcc warnings.

Thank you Joe!

PR #209, opened 2016-09-19
Set universal_newlines=True for Popen calls in hpsa and megaraid plugin utils.py

joehandzik:
This is necessary to force stdout and stderr to be strings instead of byte
sequences. The code was written assuming that was always true, which it had
been until python3 support was attempted.


tasleson:
@joehandzik Thanks!

PR #220, opened 2016-10-18
cmdline.py: Remove unnecessary whitespace

joehandzik:
While functional, there's no reason (as far as I can tell) to include these extra spaces. 


tasleson:
@joehandzik Not sure how those ended up in there, thanks!

joehandzik:
@tasleson python doesn't actually complain about them (not sure about a linter), we tested on our end and found you could add spaces, newlines, etc and not cause any problems. So we would only have seen it during a code review, but they still would have been very easy to miss.

PR #226, opened 2016-11-02
plugin/hpsa: Support Logical/Physical Block Size

### LINKED ISSUE #223: ssacli version 2.60.18.0 changes to "Logical/Physical Block Size" from "Native Block Size"
Good news is that our parser isn't totally broken now that I'm on the right unreleased versions of ssacli...bad news is that they're still changing key names on me.

joehandzik:
Newer versions of hpssacli/ssacli will output a "Logical/Physical Block Size" element rather than the previously supported "Native Block Size". Support both options.

Fixes #223 

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -651,7 +651,11 @@ class SmartArray(IPlugin):
         disk_num = key_name[len("physicaldrive "):]
         disk_name = "%s %s" % (hp_disk['Model'], disk_num)
         disk_type = _disk_type_of(hp_disk)
-        blk_size = int(hp_disk['Native Block Size'])
+        try:
+            blk_size_str = hp_disk['Native Block Size']
+        except KeyError:
+            blk_size_str = hp_disk['Logical/Physical Block Size'].split("/")[0]
+        blk_size = int(blk_size_str)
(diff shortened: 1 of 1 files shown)


joehandzik:
@tasleson @cathay4t I'll go ahead and merge this tomorrow morning if you guys don't have any other questions/comments.

tasleson:
@joehandzik I'm wondering if it would be better to test for key existence instead of using the exception path, eg.

```python
if 'Native Block Size' in hp_disk:
    blk_size_str = hp_disk['Native Block Size']
else:
    blk_size_str = hp_disk['Logical/Physical Block Size'].split("/")[0]
```
I guess I still treat exceptions as exceptional, which is kind of against the pythonic view of the world and it would execute faster, but that's not really an issue.

joehandzik:
@tasleson I'm personally indifferent, though I should probably stick with whichever way we've been doing it in the HPE plugin. So I'll take a look and if the way you suggested is more common (which I actually think it is, now that you write the code that way), I'll change this PR. Thanks!

joehandzik:
@tasleson It looks like we have several instances of using try/except to work around missing keys (see: serial number, battery status, RAID type, firmware version). So I figure we should keep this PR the way it is. Sound fair?

tasleson:
@joehandzik Sounds fair, thanks!

Issue #211, opened 2016-09-21
Feature request: Add enclosure support


BlaineEXE:
Some feedback we've gotten from one of our test labs is that we are not able to monitor the status of JBODs with lsm. If we use an lsm-only approach to management, the only way to detect JBOD health issues is to view the health LED on the physical chassis.

I've already completed about 35% of the work needed to enable this feature, but I haven't been able to work on it in a while. It would be good to develop some requirements around what details need to be included in an enclosure moving forward.

Desired properties:
- Enclosure name
- Enclosure serial number
- Enclosure link speed - Not sure if we can get this; need more research
   - 6G_SAS
   - 12G_SAS
   - OTHER
   - UNKNOWN
- Enclosure status
   - OK (healthy)
   - UNKOWN
   - FAIL (failed)
   - OTHER
- Enclosure location
   - INTERNAL (this enclosure is inside the node, e.g., Apollo internal enclosure)
   - EXTERNAL (this enclosure is outside of the node, e.g., JBOD)
   - UNKNOWN
   - OTHER
- SEPs exposed in the chassis
   - Many JBODs are physically/logically separated into multiple different SEPs. In HPE's case, the SEPs are indicative of drawer in the JBOD or backplane segment in an Apollo unit. We already have some scripts built around HPSSACLI that we would like to port to libstoragemgmt, but we need enclosure support including these SEPs to do so.
   - We also need to be able to determine which SEP a local disk belongs to to make the association

cathay4t:
@BlaineEXE I hope you mean adding more library level local APIs instead of create a plugin named as `SES` or expanding hpsa plugin. From my point view, these features you are requesting could be done via SES standard or at least SCSI T10 standards.

May I have early look on your 35% work? I used to have SES plugin created but dropped.

BlaineEXE:
The work I had done focused on using the HPSA plugin to get the information. Joe and I have discussed the SES method you mentioned, and it has merit, but I think a combination of both vendor-specific plugin and SES methods will be best. The vendor-specific plugin would be needed if one were to put a JBOD behind a RAID controller, for example. And the SES method would be preferable for a customer who knows that all controllers are going to be in HBA mode. I think this can be broken into two parts that can be developed simultaneously.

[...]

PR #237, opened 2016-12-09
Local disk link speed query

cathay4t:
C API: 
```c
int lsm_local_disk_link_speed_get(const char *disk_path,
                                  uint32_t *link_speed,
                                  lsm_error **lsm_err);
```

Python API: `lsm.LocalDisk.link_speed_get()`.

The returned link speed is in Mbps, for example, '3.0 Gbps' is `3000`.

Current, we have FC, SAS, SATA and iSCSI supported.

### DIFF
--- a/c_binding/libata.h
+++ b/c_binding/libata.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Author: Gris Ge <fge@redhat.com>
+ */
+
+#ifndef _LIBATA_H_
+#define _LIBATA_H_
+
+#include <stdint.h>
+#include "libstoragemgmt/libstoragemgmt_common.h"
+
+#define _ATA_IDENTIFY_DEVICE_DATA_LEN   512
+
+/*
+ * Preconditions:
+ *  id_dev_data != NULL
+ *  id_dev_data is uint8_t[_LIBATA_IDENTIFY_DEVICE_DATA_LEN]
+ *  link_speed != NULL;
+ * Return:
+ *  LSM_ERR_XXX.
+ */
+LSM_DLL_LOCAL int _ata_cur_speed_get(char *err_msg, uint8_t *id_dev_data,
+                                     uint32_t *link_speed);
+
+#endif  /* End of _LIBATA_H_ */
--- a/c_binding/libsg.c
+++ b/c_binding/libsg.c
@@ -35,6 +35,8 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <endian.h>
+#include <scsi/scsi.h>  /* For SCSI_IOCTL_GET_BUS_NUMBER */
+#include <limits.h>
 
 /* SGIO timeout: 1 second
  * TODO(Gris Ge): Raise LSM_ERR_TIMEOUT error for this
@@ -48,6 +50,8 @@
 /* SPC-5 rev7 Table 269 - SEND DIAGNOSTIC command */
 #define _T10_SPC_SEND_DIAG_CMD_LEN                      6
 /* SPC-5 rev7 Table 534 - Supported VPD Pages VPD page */
+#define _T10_SPC_MODE_SENSE_CMD_LEN                     10
+/* SPC-5 rev12 Table 171 - MODE SENSE(10) command */
 #define _T10_SPC_VPD_SUP_VPD_PGS_LIST_OFFSET            4
 
 /* SPC-5 rev7 4.4.2.1 Descriptor format sense data overview
@@ -164,6 +168,12 @@ struct _sg_t10_sense_dp {
     uint8_t we_dont_care_1[3];
     uint8_t len;
     /* We don't care the rest of data */
+};
+
+struct _sg_t10_mode_para_hdr {
+    uint16_t mode_data_len_be;
+    uint8_t we_dont_care_0[4];
+    uint16_t block_dp_header_len_be;
 };
 
 #pragma pack(pop)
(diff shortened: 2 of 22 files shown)


cathay4t:
@BlaineEXE Could you take a look whether this meet your expectation on issue #211 ? Thank you.

c_binding/lsm_local_disk.c
```diff
@@ -1042,3 +1128,170 @@ int LSM_DLL_EXPORT lsm_local_disk_led_status_get(const char *disk_path,
...
+    _lsm_err_msg_clear(err_msg);
+    rc = _check_null_ptr(err_msg, 3 /* argument count */, disk_path,
+                         link_speed, lsm_err);
+
+    if (rc != LSM_ERR_OK) {
+        /* set output pointers to NULL if possible when facing error in case
+         * application use output memory.
+         */
+        if (link_speed != NULL)
+            *link_speed = NULL;
```
joehandzik: @Mashimiao No it should be fine. The client shouldn't have allocated memory for us to fill. Take a look at the unit test for a usage sample.

c_binding/libsg.c
```diff
@@ -803,3 +813,130 @@ int _sg_tp_sas_addr_of_disk(char *err_msg, int fd, char *tp_sas_addr)
 
     return rc;
 }
+
+int _sg_io_mode_sense(char *err_msg, int fd, uint8_t page_code,
```
joehandzik: Is MODE SENSE (6) specifically unsupported here? When I implemented this on my end for SMART Status, I actually implemented both and had the ability to try both. If we don't need 6 I'm ok with not adding it, but curious if you know one way or the other.

c_binding/libsg.c
```diff
@@ -803,3 +813,130 @@ int _sg_tp_sas_addr_of_disk(char *err_msg, int fd, char *tp_sas_addr)
...
+                                                 * and long LBA accepted
+                                                 */
+    /* We don't need block descriptors or long LBA accepted */;
+    cdb[2] = page_code & UINT8_MAX;             /* PAGE CODE and Page control*/
+    /* The page control is 0 means 'current values' */
+    cdb[3] = sub_page_code & UINT8_MAX;         /* Subpage code */
+    cdb[4] = 0;                                 /* Reserved */
+    cdb[5] = 0;                                 /* Reserved */
+    cdb[6] = 0;                                 /* Reserved */
+    cdb[7] = (_SG_T10_SPC_MODE_SENSE_MAX_LEN >> 8 )& UINT8_MAX;
```
joehandzik: Is there a particular reason for the space after the 8 and no space between the right parenthesis and the &?

c_binding/lsm_local_disk.c
```diff
@@ -862,6 +911,43 @@ int lsm_local_disk_link_type_get(const char *disk_path,
         break;
     }
 
+    /* Use MODE SENSE(10) to query 'Protocol Specific Port mode page' as
+     * fallback.
+     */
+    if (*link_type == LSM_DISK_LINK_TYPE_NO_SUPPORT) {
+        /* Try subpage format first as hpsa return subpage format even when
```
joehandzik: I don't have any advice on this one, unfortunately...

c_binding/lsm_local_disk.c
```diff
@@ -1042,3 +1128,170 @@ int LSM_DLL_EXPORT lsm_local_disk_led_status_get(const char *disk_path,
...
+            break;
+        case _ATA_SPEED_GEN3_0:
+            *link_speed = strdup(_DISK_LINK_SPEED_6GBPS);
+            _alloc_null_check(err_msg, *link_speed, rc, out);
+            break;
+        default:
+            *link_speed = NULL;
+            rc = LSM_ERR_LIB_BUG;
+            _lsm_err_msg_set(err_msg,
+                             "BUG: Got expected ATA speed code 0x%02x",
```
joehandzik: Do you mean 'unexpected'?

python_binding/lsm/_local_disk.py
```diff
@@ -420,3 +420,34 @@ def led_status_get(disk_path):
...
+        SpecialExceptions:
+            LsmError
+                ErrorNumber.LIB_BUG
+                    Internal bug.
+                ErrorNumber.INVALID_ARGUMENT
+                    Invalid disk_path. Should be like '/dev/sdb'.
+                ErrorNumber.NOT_FOUND_DISK
+                    Provided disk is not found.
+                ErrorNumber.NO_SUPPORT
+                    Provided disk does not support SCSI SPC.
```
joehandzik: Perhaps misleading to say that this means the SCSI Spec is unsupported? Is there a more generic way to communicate that the disk does not support the spec appropriate for whatever type of device it is?

c_binding/libsg.c
cathay4t: In SCSI SPC-5 rev12:

    NOTE 20 - Implementations should migrate from the MODE SENSE(6) command to the
    MODE SENSE(10) command.

Hence I intend to no use mode sense 6 any more.

c_binding/libsg.c
cathay4t: No. I will fix it.

cathay4t:
Will upload new patch including iscsi support.

python_binding/lsm/_local_disk.py
cathay4t: No. I failed to find so.

c_binding/libata.h
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
```
tasleson: @joehandzik Sorry for the delay on answering this.  Based on this: https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility  I believe we should keep everything in the library as LGPL 2.1.  The reason I say this is LGPL 3 is not compatible with clients that are GPL2 only.  Furthermore, the nvme library that @cathay4t is working on my be an issue too as having a LGPL 2.1 library using a LGPL 3 library creates the same license restrictions.

tasleson: Just to clarify and to muddy the water, we can use different licenses if the code is executing is a different process.  Thus an out of process plugin can use a different license as it's executing in a separate process.  However, code that executes in the context of the client library should be consistently the same license.

joehandzik: @tasleson No worries, hope you had a good break! I agree with keeping everything within libstoragemgmt as LGPL 2.1, and appreciate you tracking down that compatibility FAQ. I don't think that @cathay4t's code would execute as a totally separate process from ALL libstoragemgmt code, would it? So should he change that to LGPL 2.1 too?

tasleson: @joehandzik A quick grep shows that @cathay4t has updated the copyright on the simc plugin which executes out of process and some stuff in the `c_binding` directory and the `python_binding/lsm/_clib.c` file which do not.  Those files should likely be LGPL 2.1 like the others.  Lets see what @cathay4t has to say about all this.  If we cannot get to a consensus among ourselves on what we want to do, I will consult our legal team for direction here.  In my opinion I think we should just change everything to LGPL 2.1 as it's simpler and more compatible with other licenses.

tasleson: @joehandzik Thanks for catching and raising the question of license version(s)!

joehandzik: @tasleson Sounds like a good plan. Thanks!

joehandzik: @tasleson Conversations with HPE legal in the last few years have made me paranoid. Happy to put that paranoia to some use. :)

tasleson:
From a pure API perspective, would it be better to use an enumerated value instead of a string?  Will end applications need to make run time changes based on link speed?  If yes then it would be easier to do so as an enumerated value instead of string comparisons?

c_binding/libata.h
cathay4t: @tasleson @joehandzik Thanks. I will change license to "LGPL v2.1 or later" including the NVMe library.

cathay4t:
@tasleson Enum seems better. I will add another function for converting enum to string. And I have no intention to introduce link speed changing function(too many standards and too many error possibilities).

cathay4t:
New patch set uploaded. Changes:
 * Change license to LGPL 2.1+.
 * Add iSCSI support.
 * Change returned link speed to `uint32_t`.

test/tester.c
```diff
@@ -3969,6 +3969,68 @@ START_TEST(test_local_disk_led_status_get)
...
+        lsm_err = NULL;
+    }
+
+    /* Test invalid argument */
+    rc = lsm_local_disk_link_speed_get(NULL, &link_speed, &lsm_err);
+    fail_unless(rc == LSM_ERR_INVALID_ARGUMENT,
+                "Expecting LSM_ERR_INVALID_ARGUMENT, but got %d", rc);
+    fail_unless(lsm_err != NULL, "Expecting non-NULL lsm_error, but got NULL");
+    fail_unless(link_speed == LSM_DISK_LINK_SPEED_UNKNOWN,
+                "Expecting link_speed been set as LSM_DISK_LINK_SPEED_UNKNOWN, "
```
joehandzik: All of these "Expecting link_speed been set..." error messages in this file should be changed to something like "Expecting link_speed to have been set...". It'll make the error messages easier to read.

c_binding/utils.c
```diff
@@ -3,7 +3,7 @@
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
  * License as published by the Free Software Foundation; either
- * version 3 of the License, or (at your option) any later version.
+ * version 2.1 of the License, or (at your option) any later version.
```
joehandzik: For any of the files that had previously been changed to LGPL v3, I'd rather see a separate PR to fix that issue. And we should get that committed ASAP. So, just cherry-pick this patch out of this PR and create a new one.

cathay4t:
@joehandzik The grammar of C unit test error message has been updated.
PR #241 is for license change.

joehandzik:
@cathay4t Looks like you still have a the 2.1 reversion in the utils.c/h files, but a rebase with master should clear that up. Otherwise, LGTM.

tasleson:
@cathay4t the functions:

`_fc_host_speed_get` and `_iscsi_host_speed_get` are very similar.  Would it be possible to reduce this to one common code path for both?

cathay4t:
@tasleson Merged `_fc_host_speed_get` and `_iscsi_host_speed_get`.

New patch uploaded.

tasleson:
@cathay4t Memory leak?

```
======== /tmp/29743/logs//mem_leak_leaking_client BEGIN ==========
==21631== 57 (56 direct, 1 indirect) bytes in 1 blocks are definitely lost in loss record 6 of 7
==21631==    at 0x4C29DB4: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==21631==    by 0x4E4F913: lsm_error_create (lsm_datatypes.cpp:338)
==21631==    by 0x4E8887C: lsm_local_disk_list (lsm_local_disk.c:836)
==21631==    by 0x4065E2: test_local_disk_led_status_get (tester.c:3929)
==21631==    by 0x41B0B2: srunner_run_all (in /tmp/29743/bin/tester)
==21631==    by 0x405FAC: main (tester.c:4113)
```

cathay4t:
Checking the code. Wired it pop up in this PR.

cathay4t:
New patch included fixing the memory leak found by CI.

tasleson:
@cathay4t Thank you!

PR #241, opened 2017-01-10
Change license to "LGPL v2.1 or later".

cathay4t:
According to https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility
LGPL v3.0+ is not compatible with GPL 2.0 only application.
Hence we relicense to LGPL v2.1+ for better compatibilities.


joehandzik:
@cathay4t @tasleson LGTM, gonna pull this in despite the hitachi plugin not being done...no idea what would gum it up, but I expect it to be independent of the changes here.

PR #242, opened 2017-01-17
Remove glib dependency.

cathay4t:
```
Why:

    * glib exit when no memory while lsm as a library just return `LSM_ERR_NO_MEMORY`;

    * glib is too heavy for this project IMHO.

How:
    * Use utlist to replace GPtrArray.
    * Use uthash to replace GHashTable which is used by 'lsm_hash'.

```


cathay4t:
Please hold on, performance test on 1K level seem unfair, let me run performance test on 10k level.

cathay4t:
Got some converity scan error also, will fix soon.

tasleson:
>Why:
>
>   * glib exit when no memory while lsm as a library just return `LSM_ERR_NO_MEMORY`;
>
>    * glib is too heavy for this project IMHO.

In general with linux's opportunistic memory handling, it could very well return a valid pointer only to find out later that it cannot really full-fill the request.  As for glib being too heavy, yeah maybe, but in general I like using other code that is well tested and supported by someone else :-)

All that being said, I would be more accepting of integrating this change if it was a drop in replacement for the glib one.  It should be possible to mimic the same API as glib and then use an ifdef to switch from one to the other.  That way if we ran into a bug before or during us trying to get a release out, we could easily toggle from one to the other.  A configure option to do that isn't required, but it would be nice :-)

cathay4t:
@tasleson Thanks. I will use utlist and uthash for this replacement and wrap it into the same API of glib.

joehandzik:
@cathay4t Sorry for the radio silence, been jockeying between projects a bit lately. I think your latest comment about making this configurable as an option is the way to go. I personally haven't run into any problems with glib in our use at HPE, FWIW.

cathay4t:
@tasleson New version uploaded:
 * Use utlist.h and uthash.h for pointer array and hash table.
 * `configure --enable-glib` could switch back to old glib implementation.

cathay4t:
Please hold on. I got a memory leak in glib mode.

cathay4t:
Memory leak has been fixed. Please review. Thanks.

tasleson:
``` patch
diff --git a/configure.ac b/configure.ac
index 2d8e06d..e21ec6b 100644
--- a/configure.ac
+++ b/configure.ac
@@ -345,7 +345,8 @@ AX_CHECK_COMPILE_FLAG([-Wall], [CFLAGS="$CFLAGS -Wall"], [], [])
 AX_CHECK_COMPILE_FLAG([-Werror], [CFLAGS="$CFLAGS -Werror"], [], [])
 AX_CHECK_COMPILE_FLAG([-Wextra], [CFLAGS="$CFLAGS -Wextra"], [], [])
 
-AC_ARG_ENABLE(glib, "Use glib for hash and pointer array",
+AC_ARG_ENABLE([glib],
+              [AS_HELP_STRING([--enable-glib], [Use glib for hash and pointer array])],
               [use_glib=yes], [])
 if test "x$use_glib" == "xyes"; then
     PKG_CHECK_MODULES([LIBGLIB], [glib-2.0 >= 2.22.5])
```
The output of `./configure --help` doesn't look correct, the above change makes things look as expected.

tasleson:
@cathay4t covscan results, please check out

```
Error: CLANG_WARNING: [#def3]
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1759:20: warning: Potential leak of memory pointed to by 'k_value'
#            return LSM_ERR_OK;
#                   ^
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1753:5: note: Taking true branch
#    if (LSM_IS_HASH(op)) {
#    ^
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1754:19: note: Memory is allocated
#        k_value = strdup(key);
#                  ^~~~~~~~~~~
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1757:13: note: Left side of '&&' is true
#        if ((k_value != NULL) && (d_value != NULL) &&
#            ^
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1757:13: note: Left side of '&&' is true
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1757:9: note: Taking true branch
#        if ((k_value != NULL) && (d_value != NULL) &&
#        ^
libstoragemgmt-1.5.0/c_binding/lsm_datatypes.cpp:1759:20: note: Potential leak of memory pointed to by 'k_value'
#            return LSM_ERR_OK;
#                   ^
# 1757|           if ((k_value != NULL) && (d_value != NULL) &&
# 1758|               (_hash_table_set(op->data, k_value, d_value) == 0))
# 1759|->             return LSM_ERR_OK;
# 1760|           else
# 1761|               goto nomem;

Error: CONSTANT_EXPRESSION_RESULT (CWE-398): [#def4]
libstoragemgmt-1.5.0/c_binding/ptr_array_glib.c:42: logical_vs_bitwise: The expression "size && 1 /* 2147483647 * 2U + 1U */" is suspicious because it performs a Boolean operation on a constant other than 0 or 1.
libstoragemgmt-1.5.0/c_binding/ptr_array_glib.c:42: remediation: Did you intend to use "&" rather than "&&"?
#   40|           return NULL;
#   41|       }
#   42|->     pa->array = g_ptr_array_sized_new((guint) (size && G_MAXUINT));
#   43|       return pa;
#   44|   }
```

cathay4t:
@tasleson Thanks for the reviews. New patches uploaded. For the `Potential leak of memory`, it's a false alarm, I carefully re-read the code and still not found the memory leak issue. If you try covscan with `--profile=cov-unstable`, it does not report this memory leak issue.

tasleson:
@cathay4t Curious to why a few files have the author email address as your gmail address?

tasleson:
@cathay4t  From looking at https://troydhanson.github.io/uthash/userguide.html it would appear that you should be using `HASH_ADD_KEYPTR` instead of `HASH_ADD_STR`?

c_binding/hash_table_glib.c
```diff
@@ -0,0 +1,174 @@
...
+        free(*keys);
+        goto nomem;
+    }
+
+    /* initialize the array to all NULL */
+    for (i = 0; i < *count; ++i) {
+        (*keys)[i] = NULL;
+        (*values)[i] = NULL;
+    }
+
```
tasleson: How about using a couple of `memset` instead?

jumitche: Use calloc() instead, it does the malloc and memset in one operation and removes the multiply.

cathay4t: Thanks. Will use `calloc()`.

cathay4t:
@tasleson I use vim function to auto-generate license lines, I may used the incorrect one for the auther email. Will fix it.
Will check `HASH_ADD_KEYPTR` also.

c_binding/hash_table_glib.c
tasleson: 👍

cathay4t:
The performance of using utlist replacing glib pointer array seems not a good idea:

```
# glib ptr array for 10,000 string.

real    0m0.125s
user    0m0.100s
sys     0m0.007s

# utlist fro 10,000 string

real    0m1.350s
user    0m1.310s
sys     0m0.013s
```

cathay4t:
@tasleson Any input on which I should use on replacing glib pointer array we used for `lsm_string_list`:
 * A: Use self-implemented pointer array. (that's what I did initially in this PR).
 * B: We use utlist which is well-tested(I guess) with performance step back.

tasleson:
@cathay4t If you google you find all kinds of suggestions to utilize glib :-).  I really don't want to see us writing our own container functionality for maps & vectors.  The only other thought I have is using C++ STL stuff.  Internally we are using C++ so we could leverage that, although at one point we were trying to migrate away from using C++ too.

cathay4t:
@tasleson I believe I could call this PR a good lesson and leave glib and c++ as it were in libstoragemgmt until we found real problem or real improvement.

PR #249, opened 2017-02-22
udev: Fix gcc warning on non-x86 platform.

cathay4t:
* Issue:
    Got failure on s390x:

        scan-scsi-target.c:90:2: error: comparison is always true due to limited
        range of data type [-Werror=type-limits]

* Root cause:
    The error is on these lines:

        char c;
        ...
        while ((c = getopt_long(argc, argv, "rh", longopts, NULL)) != -1) {

  On non-x86 platform, char is unsigned. Which makes the (c != -1)
  always true.

* Fix:
    Take return of getopt_long() as int.


tasleson:
:+1:  LGTM  Thanks Gris!

PR #254, opened 2017-02-27
HPSA plugin: Workaround more changes to ssacli format

joehandzik:

Caught by @cvedel when he ran an updated version of ssacli. Note that I am trying to make it more difficult for Microsemi to get away with making these inane changes, but it has been slow-going. Hopefully this set of bug reports with encourage them to stop and/or provide the structured output that I've been requesting for two years.


joehandzik:
The RAID failures were a false positive (sort of), the problem was actually with the version of ssacli that I was running on the test node. I needed to move to ssacli 2.65.7.0 to get volume LED enablement to work.

PR #261, opened 2017-03-28
Update trusted.yaml to include libstorage

joehandzik:

Seems like we should trust the root repo...

### DIFF
--- a/test/trusted.yaml
+++ b/test/trusted.yaml
@@ -12,3 +12,4 @@ REPOS:
 - "https://github.com/Mashimiao/libstoragemgmt.git"
 - "https://github.com/jumitche/libstoragemgmt.git"
 - "https://github.com/cvedel/libstoragemgmt.git"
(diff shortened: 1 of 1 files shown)


cathay4t:
@joehandzik Any particular reason for using branches in root repo? Messing with branches in root repo directly might confuse people and maloperation(like `git push origin +master` might cause problem.

joehandzik:
@cathay4t I guess it's more up to you guys than me, there are just cases where it's easier to generate a branch from a master checkout than from a personal fork checkout, but it can be worked around. We can avoid this if you and @tasleson really want to.

tasleson:
@joehandzik @cathay4t  I'm OK either way, we should update the trusted file to include master if we want CI to run.

tasleson:
The beauty of git is we can always fix stuff if we mess up :-)

tasleson:
Case in point :-)

PR #260, opened 2017-03-28
HPSA: Work around Smart HBA name mismatch

joehandzik:
Signed-Off-By: Joe Handzik <joseph.t.handzik@hpe.com>

Fixes #258

CI Kick

PR #262, opened 2017-03-28
libsg: Fix fixed sense data structure

### LINKED ISSUE #259: Structure for fixed sense data is incorrect in libsg
It should actually be one byte longer for the first "don't care" array. Found out because I actually do care for some of my SMART Status code. Patch incoming.

joehandzik:
There are four bytes in the 'Information' portion of the structure, not three.

Fixes #259


joehandzik:
The failure is expected on my node until #260 is merged.

cathay4t:
Thanks. Merged.

PR #271, opened 2017-06-13
New test cases for Python 2 and Python3 interaction test.

cathay4t:
Test: Two new test cases for Python 2 and 3 interaction.
    
     * Run python2 version of lsmcli test and plugin test with python3
       version simulator plugin.
    
     * Run python3 version of lsmcli test and plugin test with python2
       version simulator plugin.
    
     * These two tests will be run only when `--with-python3` configure
       option is in use.

Also fixed some issue found by these tests.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,6 +193,7 @@ AC_ARG_WITH([python3],
     [])
 
 if test "x$with_python3" = "xyes"; then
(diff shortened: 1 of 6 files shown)


tasleson:
@cathay4t Thanks Gris

PR #272, opened 2017-06-15
lsmcli: Fix bug of fs-snap-restore when restoring the whole file system

cathay4t:
```
Issue:
    Got error when `lsmcli fs-snap-restore --force  --fs <FS_ID> --snap <SNP_ID`
    against a NetApp ONTAP array:
        INVALID_ARGUMENT(101): Invalid parameter combination 

Root cause:
    In API definition, only when "files == None and all_files = True"
    means all files on the file system are restored. The lsmcli set the
    `files` to [] which is None.

Fix:
    Fix lsmcli to use None for restoring the whole file system.
```


tasleson:
Thanks @cathay4t

PR #273, opened 2017-06-23
Sim and simc plugin: Fix missing disks after volume_delete() on HW RAID volume

cathay4t:
```
Issue:
    When deleting a volume created by `volume_raid_create()`, disks used
    by this volume will also be deleted.

Root cause:

    The `disks` table will remove the disks entry when owner pool delete.
    The sql table line to blame is:

            FOREIGN KEY(owner_pool_id)
                REFERENCES pools(id) ON DELETE CASCADE );

Fix:

    * Change to `ON DELETE SET DEFAULT`.
    * Update disk `role` to NULL when deleting a HW RAID volume.

```

### DIFF
--- a/plugin/sim/simarray.py
+++ b/plugin/sim/simarray.py
@@ -129,7 +129,7 @@ class PoolRAID(object):
 
 
 class BackStore(object):
-    VERSION = "4.0"
+    VERSION = "4.1"
     VERSION_SIGNATURE = 'LSM_SIMULATOR_DATA_%s_%s' % (VERSION, md5(VERSION))
     JOB_DEFAULT_DURATION = 1
     JOB_DATA_TYPE_VOL = 1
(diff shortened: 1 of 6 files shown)


plugin/sim/simarray.py
```diff
@@ -232,7 +232,7 @@ def __init__(self, statefile, timeout):
             rpm INTEGER,
             link_type INTEGER,
             FOREIGN KEY(owner_pool_id)
-            REFERENCES pools(id) ON DELETE CASCADE );
+            REFERENCES pools(id) ON DELETE SET DEFAULT);
```
Mashimiao: Here, should be `ON DELETE SET NULL`? As, owner_pool_id seems doesn't have default value.

cathay4t: The default is NULL: http://www.sqlite.org/lang_createtable.html

> The default value of each column is NULL.

Mashimiao: OK, got it

Mashimiao:
LGTM

tasleson:
@cathay4t Please add a unit test which reproduces this issue.

cathay4t:
@tasleson Plugin test case added for this issue.

tasleson:
Thanks @cathay4t

PR #276, opened 2017-06-29
Refactored as per the review comments

brraghav:
Removed the parse_config_output as JSON parser is available now.


brraghav:
Need to resolve conflict, so removing the PR.

PR #278, opened 2017-06-29
Removed the arcconf parser and using only the JSON parser.

brraghav:
Removed the old arcconf parser and having a single JSON parser.
Added a new method to get the number of controllers instead.


brraghav:
Added the changes to the old PR itself. Hence closing.

PR #281, opened 2017-07-11
Improve targetd plugin SSL certificate handling

tasleson:
Two new URI parameters have been added:

`no_ssl_verify=yes` = Disables all SSL certificate validation
`cert_file=<server certificate file >` = Certificate to use to verify server certificate

Targetd man page updated as well.

### DIFF
--- a/plugin/targetd/targetd.py
+++ b/plugin/targetd/targetd.py
@@ -22,6 +22,7 @@ import socket
 import re
 import base64
 import six
+import ssl
 
 from lsm import (Pool, Volume, System, Capabilities,
                  IStorageAreaNetwork, INfs, FileSystem, FsSnapshot, NfsExport,
@@ -121,6 +122,7 @@ class TargetdStorage(IStorageAreaNetwork, INfs):
         self.scheme = None
         self.url = None
         self.headers = None
+        self.no_ssl_verify = False
         self._flag_ag_support = True
         self.system = System("targetd", "targetd storage appliance",
                              System.STATUS_UNKNOWN, '')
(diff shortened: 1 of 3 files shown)


plugin/targetd/targetd.py
```diff
@@ -148,6 +152,20 @@ def plugin_register(self, uri, password, timeout, flags=0):
         self.headers = {'Content-Type': 'application/json',
                         'Authorization': 'Basic %s' % (auth,)}
 
+        if "no_ssl_verify" in self.uri["parameters"] \
+                and self.uri["parameters"]["no_ssl_verify"] == 'yes':
+            self.no_ssl_verify = True
+
+        if "cert_file" in self.uri["parameters"]:
+            # Check for file existence and throw error now if not present
+            self.cert_file = self.uri["parameters"]["cert_file"]
```
cathay4t: Can we raise `INVALID_ARGUMENT` error for when both `no_ssl_verify` and `cert_file` is used?
`targetd+ssl://localhost?cert_file=<file_path>&no_ssl_verify=yes`

plugin/targetd/targetd.py
```diff
@@ -148,6 +152,20 @@ def plugin_register(self, uri, password, timeout, flags=0):
...
+
+            if not (os.path.exists(self.cert_file) and
+                    os.path.isfile(self.cert_file)):
```
cathay4t: `os.path.isfile` also check the existence: https://docs.python.org/2/library/os.path.html#os.path.isfile

cathay4t:
The `create_default_context` is new in python 2.7.9, we might need to enforce that in `configure.ac` or implement a workaround for python 6 or raise `INVALID_ARGUMENT` when `cert_file` is used in python 2.6 system.

tasleson:
@cathay4t Well only the newer versions of python probably do the certificate checking, so we should be able to maintain the old behavior, but I'll see what I can do to enable the same new features across all supported versions.

tasleson:
For newer versions of python, all our plugins that use urllib/urlib2 library probably need the same set of changes to support disable and adding cert file.

plugin/targetd/targetd.py
tasleson: Done

tasleson:
@cathay4t I've added the suggestions you made.  I also cleaned up the error handling for connection based errors.  Testing on F25 (py3) and EL7 & 6 (py2)

cathay4t:
Code looks good.
All manual tests passed:
 * `no_ssl_verify=yes`
 * `cert_file` conflicting with `no_ssl_verify`
 * `cert_file` with incorrect hostname.
 * Correct `cert_file`.

cathay4t:
Merged. Thank you Tony!

PR #284, opened 2017-07-20
C library: Fix bit field endian compatibility issue

cathay4t:
```
Issue:
    C local disks functions does not work on big-endian
    system(s390x/ppc64/arm/etc).

Root cause:
    Bit field has endian compatibility issue.
    https://stackoverflow.com/questions/6043483/why-bit-endianness-is-an-issue-in-bitfields

Fix:
    Use gcc macro to detect endianness and reorder the struct member
    sequence. It's also the way used by linux kernel tcp module:
        /usr/include/linux/tcp.h
```

### DIFF
--- a/c_binding/libses.h
+++ b/c_binding/libses.h
@@ -39,6 +39,7 @@
 struct _ses_dev_slot_status {
     uint8_t common_status;
     uint8_t diff_between_dev_slot_and_array_dev_slot;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
     uint8_t report              : 1;
     uint8_t ident               : 1;
     uint8_t rmv                 : 1;
@@ -47,6 +48,7 @@ struct _ses_dev_slot_status {
     uint8_t enc_bypass_a        : 1;
     uint8_t do_not_remove       : 1;
     uint8_t app_bypass_a        : 1;
+
     uint8_t dev_bypass_b        : 1;
     uint8_t dev_bypass_a        : 1;
     uint8_t bypass_b            : 1;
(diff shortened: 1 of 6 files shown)


cathay4t:
@joehandzik Just FYI since you might be working on new local disk functions.

joehandzik:
@cathay4t Good find, and thanks for the heads up! Now to figure out why my HPE boxes aren't showing up in CI...

cathay4t:
`make check` works well on ppc64 RHEL 7.

cathay4t:
@tasleson Unfortunately, gcov/lcov does not mark line of struct definition as covered, hence we cannot
tell from gcov whether the `__BYTE_ORDER__` line works or not. But from manual test on ppc64 server,
it works well.

cathay4t:
And the CI is running on ppc64 RHEL 7.4 server using `sudo make check` command which covers all the
`LocalDisk` functions.

c_binding/libses.h
```diff
@@ -55,6 +57,25 @@ struct _ses_dev_slot_status {
     uint8_t fault_reqstd        : 1;
     uint8_t fault_sensed        : 1;
     uint8_t app_bypass_b        : 1;
+#else
+    uint8_t app_bypass_a        : 1;
+    uint8_t do_not_remove       : 1;
+    uint8_t enc_bypass_a        : 1;
+    uint8_t enc_bypass_b        : 1;
+    uint8_t ready_to_insert     : 1;
```
tasleson: This one isn't present on little endian?

c_binding/libses.h
```diff
@@ -55,6 +57,25 @@ struct _ses_dev_slot_status {
     uint8_t fault_reqstd        : 1;
     uint8_t fault_sensed        : 1;
     uint8_t app_bypass_b        : 1;
+#else
+    uint8_t app_bypass_a        : 1;
+    uint8_t do_not_remove       : 1;
+    uint8_t enc_bypass_a        : 1;
+    uint8_t enc_bypass_b        : 1;
```
tasleson: This one isn't present for little endian?

c_binding/libses.h
```diff
@@ -55,6 +57,25 @@ struct _ses_dev_slot_status {
...
+
+    uint8_t app_bypass_b        : 1;
+    uint8_t fault_sensed        : 1;
+    uint8_t fault_reqstd        : 1;
+    uint8_t dev_off             : 1;
+    uint8_t bypass_a            : 1;
+    uint8_t bypass_b            : 1;
+    uint8_t dev_bypass_a        : 1;
+    uint8_t dev_bypass_b        : 1;
+#endif
```
tasleson: Little endian has 12 bits, big has 16, need some padding?

c_binding/libses.h
cathay4t: Are you mean `enc_bypass_b`?
It is in little endian definition in reverse order.

c_binding/libses.h
cathay4t: Are you mean `ready_to_insert`?
It is in little endian definition in reverse order.

c_binding/libses.h
cathay4t: Both are 16 bits. I added new empty line at line 51 to make sure I didn't miss anything.

tasleson: Hmmm, if I look at the full file I see them, I think the web diff wasn't showing them which was confusing.

tasleson:
@cathay4t  Thanks Gris!

PR #290, opened 2017-08-08
 LocalDisk: Fix error on Dell virtual floppy

cathay4t:
```
Issue:

    Got this error of command `lsmcli ldl` against a Dell virtual
    floppy:

        LIB_BUG(1): BUG: Got incorrect VPD page code '80', should be 0x83

Root cause:

    When query VPD83 of Dell virtual floppy, it returns a STANDARD
    INQUIRY instead of VPD page, and this is not clarified by SCSI SPC-4
    or later version.

    [fge@dell-pe1950-06 ~]$ sudo sg_vpd --page=0x83 /dev/sdc -vvv -H
    open /dev/sdc with flags=0x800
    Device Identification VPD page:
        inquiry cdb: 12 01 83 00 fc 00
          duration=1 ms
        inquiry: pass-through requested 252 bytes but got 36 bytes
    invalid VPD response; probably a STANDARD INQUIRY response
    First 32 bytes of bad response
     00     00 80 00 01 1f 00 00 00  44 65 6c 6c 20 20 20 20    ........Dell
     10     56 69 72 74 75 61 6c 20  20 46 6c 6f 70 70 79 20    Virtual  Floppy
    fetching VPD page failed

Fix:
    Treat it as no support.
    This malformed data might fit into VPD 80 data, but the VPD 0x80
    data taken by _sg_parse_vpd_80() is retrieved via sysfs where Linux
    kernel already checked the support status of VPD pages, hence this
    will not trigger unknown action of _sg_parse_vpd_80() due to
    malformed scsi data.
```

### DIFF
--- a/c_binding/libata.c
+++ b/c_binding/libata.c
@@ -40,10 +40,16 @@
 
 #pragma pack(push, 1)
 struct _ata_sata_add_cap {
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
     uint8_t zero            : 1;
     uint8_t cur_speed       : 3;
-    uint8_t we_dont_care_0  : 8;
-    uint8_t we_dont_care_1  : 4;
+    uint8_t we_dont_care_0  : 4;
+#else
+    uint8_t we_dont_care_0  : 4;
+    uint8_t cur_speed       : 3;
+    uint8_t zero            : 1;
+#endif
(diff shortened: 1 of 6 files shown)


cathay4t:
Please merge #284  before this one. Thanks.

tasleson:
@cathay4t Thanks Gris!

PR #293, opened 2017-08-22
NFS plugin: Change system id and name.

cathay4t:
* Change system id and name to include NFS in it.


cathay4t:
Since there will be an `local` plugin, using `local` as system ID might be confusing.

@jumitche Does this change works for you?

cathay4t:
Merged due to no response and no harm.

PR #294, opened 2017-08-22
Fix assert bug in _fill_ata_regs_fixed

joehandzik:
Clearly this path was never used in my testing. Checking the correct variable for assert.


joehandzik:
I went ahead and merged this since it's clearly a bug with a one line fix.

cathay4t:
@joehandzik Sure. Thanks.

Issue #291, opened 2017-08-17
One bad disk response causes local-disk-list to fail entirely


joehandzik:
Logging this here in case anyone has a different opinion, but I was walking someone through using libstoragemgmt on a Gen10 ProLiant, and we ran into an unhandled return code on his box for one of the operations executed in local-disk-list. Unfortunately, that prevents all further operations from executing. So, we'll want to do something else other than raising the errors that come out from operations in local-disk-list, IMO.

cathay4t:
@joehandzik Good point.
How about we change lsmcli only print a error line instead of falling the whole command?

joehandzik:
@cathay4t That's basically what I've done as a workaround. I should have self-assigned this, since I have something complete already. Just working through another bug before submitting a PR.

tasleson:
Was the error in retrieving the list of local disks or when we were gathering more information on each disk?

joehandzik:
@tasleson Gathering information phase. We think we actually tracked it down to an SD card that was hitting an area where we passed up an error instead of just returning NO_SUPPORT.

cathay4t:
Closing since fix was merged.

PR #300, opened 2017-08-31
C library: Switch from sg kernel module to bsg.

cathay4t:
* sg kernel module is not auto-loaded on some(RHEL/Centos 7)
   distribution, while bsg is mostly compiled(RHEL/Fedora/Ubuntu) into kernel.
   So, we use /dev/bsg/<htbl_id> instead of /dev/sgX for SES functions.

 * The bsg only support SG_IO v4(while sg only for V3), hence we have
   new function `_sg_io_v4()` for that.

 * Also ignore error of `lsmcli ldl`, fix issue #291.

### DIFF
--- a/tools/lsmcli/cmdline.py
+++ b/tools/lsmcli/cmdline.py
@@ -1822,7 +1822,9 @@ class CmdLine(object):
                     info_dict[key] = func_dict[key](disk_path)
                 except LsmError as lsm_err:
                     if lsm_err.code != ErrorNumber.NO_SUPPORT:
-                        raise
+                        sys.stderr.write("WARN: %s('%s'): %d %s" %
+                                         (func_dict[key].__name__, disk_path,
+                                          lsm_err.code, lsm_err.msg))
 
             local_disks.append(
                 LocalDiskInfo(disk_path,
--- a/c_binding/libses.c
+++ b/c_binding/libses.c
@@ -35,11 +35,11 @@
 #include <assert.h>
 
 /* SPC-5 Table 139 - PERIPHERAL DEVICE TYPE field */
-#define _LINUX_SG_DEV_TYPE_SES                  "13"  /* 0x0d  */
-#define _LINUX_SG_DEV_TYPE_SES_LEN              2
+#define _LINUX_SCSI_DEV_TYPE_SES                "13"  /* 0x0d  */
+#define _LINUX_SCSI_DEV_TYPE_SES_LEN            2
 /* just two digits like above */
 
-#define _SYSFS_SG_ROOT_PATH                     "/sys/class/scsi_generic"
+#define _SYSFS_BSG_ROOT_PATH                    "/sys/class/bsg"
 
 #define _T10_SES_CFG_PG_CODE                    0x01
 #define _T10_SES_STATUS_PG_CODE                 0x02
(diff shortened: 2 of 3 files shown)


cathay4t:
@joehandzik 
 * I fixed issue #291 which I believe you is working on. Let me know if that's not what you are seeking.
 * This PR changed from sg kernel module to bsg, could you test the SES functions on your HPE servers?

Thanks.

tools/lsmcli/cmdline.py
```diff
@@ -1822,7 +1822,9 @@ def local_disk_list(self, args):
                     info_dict[key] = func_dict[key](disk_path)
                 except LsmError as lsm_err:
                     if lsm_err.code != ErrorNumber.NO_SUPPORT:
-                        raise
+                        print("BUG: %s('%s'): %d %s" %
```
joehandzik: Calling errors here a bug is a little overdoing it from my perspective. It'll make the customer think something is wrong with the software, even if the errors is with unsupported hardware (like an SD card, which we ran into). I would simply cut off the "BUG: " portion of your current message.

joehandzik:
@cathay4t I'll check as soon as I can, but our lab hardware is offline due to the Houston flooding. It may be a full month before I can check. :(

cathay4t:
@joehandzik Sure. This is PR is not that hurry.

tasleson:
@cathay4t 

```
libsg.c: In function ‘_sg_io_v4’:
libsg.c:422:22: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
     io_hdr.request = (__u64) cdb;
                      ^
libsg.c:423:23: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
     io_hdr.response = (__u64) sense_data;
                       ^
libsg.c:429:32: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
             io_hdr.din_xferp = (__u64) data;
                                ^
libsg.c:432:33: error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
             io_hdr.dout_xferp = (__u64) data;
```

cathay4t:
@tasleson The compile works on all my environments. On which platform/distribution did you get this error?.

tasleson:
|The compile works on all my environments. On which platform/distribution did you get this error?.
That was from the Snia lab systems, click the Details link on the CI results.  So fedora something, need to check which version.

cathay4t:
Patch tested on RHEL 7.4 for SES enclosure functions.

cathay4t:
The RHEL 6 compiling pass after remove the AX_CHECK_COMPILE_FLAG lines and a unrelated nfs C lib warning.

tools/lsmcli/cmdline.py
cathay4t: I changed the word "BUG" to "WARN" in new patch.

tools/lsmcli/cmdline.py
```diff
@@ -1822,7 +1822,9 @@ def local_disk_list(self, args):
                     info_dict[key] = func_dict[key](disk_path)
                 except LsmError as lsm_err:
                     if lsm_err.code != ErrorNumber.NO_SUPPORT:
-                        raise
+                        print("WARN: %s('%s'): %d %s" %
```
tasleson: The only bad thing about doing this is if anyone is scripting lsmcli this will break them.  I wonder if we shouldn't print this on stderr instead of stdout?  At least then a user could sort out what they want to parse.

tasleson:
@cathay4t What hardware has this been tested on?  eg. 32/64 bit x86, big/little endian etc.

tasleson:
@cathay4t Please take a peek at the NetApp setup too, something is amiss there.

cathay4t:
@tasleson I believe the Netapp CI failure is caused by our QE are rebooting ONTAP filer in the middle of our tests.

I tested the real SES functions on Fedora 26 x86_64 and RHEL 7.4 x86_64.

tools/lsmcli/cmdline.py
cathay4t: Changed to stderr.

tasleson:
@cathay4t Thanks Gris!

PR #306, opened 2017-09-25
Enable GCC warnings on CPP code also.

cathay4t:
* Change configure.ac to set gcc warnings flags in CPPFLAGS.
 * Fix all warnings.


tasleson:
@cathay4t Looks good

PR #307, opened 2017-09-25
Improve pool status of MegaRAID plugin

cathay4t:
* Support `Pool.STATUS_RECONSTRUCTING` and `Pool.STATUS_VERIFYING`. Tested on MegaRAID 9286CV-8eCC
 * Fix when both `perccli` and `storcli` are installed. Tested on Dell PERC and Broadcom MegaRAID

### DIFF
--- a/plugin/megaraid/megaraid.py
+++ b/plugin/megaraid/megaraid.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2015-2016 Red Hat, Inc.
+# Copyright (C) 2015-2017 Red Hat, Inc.
 # This library is free software; you can redistribute it and/or
 # modify it under the terms of the GNU Lesser General Public
 # License as published by the Free Software Foundation; either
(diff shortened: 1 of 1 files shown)


tasleson:
@cathay4t Changes look good, I did see something non related to these changes which we should consider addressing:

``` python
@_handle_errors
def plugin_register(self, uri, password, timeout, flags=Client.FLAG_RSVD):
    if os.geteuid() != 0:
        raise LsmError(
            ErrorNumber.INVALID_ARGUMENT,
            "This plugin requires root privilege both daemon and client")
    uri_parsed = uri_parse(uri)
    self._storcli_bin = uri_parsed.get('parameters', {}).get('storcli')
    # change working dir to "/tmp" as storcli will create a log file
    # named as 'MegaSAS.log'.

    os.chdir("/tmp")
    if self._storcli_bin:
        self._storcli_exec(['-v'], flag_json=False)
    else:
        self._find_storcli()
```

It's bad to have a file with a known name and location that someone can predict to exist in /tmp from a security standpoint.  A person could link to a system file etc and depending on what the command line does it could unintentionally trash/delete a system file or possibly some other creative way to create havok.  One suggestion would be to use one of the secure functions to create a random directory in /tmp with permissions restricted to owner and use that as the base directory to do the `os.chdir` into.

cathay4t:
@tasleson That's a good idea. I will use the random folder for this log file. Thanks.

tasleson:
@cathay4t Thanks Gris

PR #308, opened 2017-09-27
Improve SSL communication on plugins.

cathay4t:
* SMIS-S plugin now support pass CA file via URI parameter `ca_cert_file`. Tested against EMC SMI-S provider.
 * Targetd plugin: Rename URI parameter 'cert_file' to 'ca_cert_file'
 * ONTAP plugin now support pass CA file via URI parameter `ca_cert_file`. Tested on RHEL 7[1].

[1]: Fedora treat ONTAP 8 TLS version 1 cipher `DES-CBC3-SHA` as week and no pass. Archlinux completely removed the support of that cipher.


tasleson:
@cathay4t Thanks Gris, the failures in the CO lab are caused by missing python3 pywbem.  However, at the moment I can't get anything to install on that box because `rpm` seems to be borked some how.

cathay4t:
@tasleson You may find a working server with same OS version and scp the files of `rpm` there or just reinstall OS.

PR #309, opened 2017-09-28
MegaRAID plugin: Use random temperate folder for log of storlci.

cathay4t:
Background story is:
https://github.com/libstorage/libstoragemgmt/pull/307#issuecomment-332312221

This patch use random temperate folder to store the log of
storcli/perccli.

tasleson: CI Kick

### DIFF
--- a/plugin/megaraid/megaraid.py
+++ b/plugin/megaraid/megaraid.py
@@ -19,6 +19,8 @@ import json
 import re
 import errno
 import math
+import tempfile
+import shutil
 
 from lsm import (uri_parse, search_property, size_human_2_size_bytes,
                  Capabilities, LsmError, ErrorNumber, System, Client,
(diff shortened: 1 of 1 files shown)


tasleson:
@cathay4t This change looks pretty good, however I don't think we can leverage `__del__` to ensure that the temporary directory gets removed.  The `__del__` method only gets run on garbage collection which is not guaranteed to occur.

I wonder if we should leverage the `atexit` or something else to ensure we clean up after ourselves?

cathay4t:
@tasleson Thanks. I am reading https://docs.python.org/3.6/library/weakref.html#comparing-finalizers-with-del-methods which is exactly our case here.

cathay4t:
@tasleson I'd like to keep using the `__del__` way:
 * The weakref way mentioned by python3 page does not works on python2.
 * The `atexit` way has issue: if `lsm.Client('megaraid://')` raised an exception, the temp folder only be deleted when next call of lsm.Client('megaraid://') in interactive mode of python2.

And I am little confused/irritated by python on `do not guarantee`. Anyway idea which way we should take?

tasleson:
@cathay4t Does using `__del__` appear to work reliably?  I'm under the impression that it won't even get called unless garbage collection is called, which may not happen.  Perhaps we just need to put it in one of the exit paths and during normal code execution it will get cleaned up.  I don't think we can guarantee it gets removed, but we should make best effort to do so.

cathay4t:
@tasleson I tested the `__del__` on exception, connection close, daemon quit. All works well.

tasleson:
@cathay4t Thanks Gris!

PR #311, opened 2017-09-30
Trivial bug fixes.

cathay4t:


c_binding/libsg.c
```diff
@@ -412,7 +412,7 @@ static int _sg_io_v4(int fd, uint8_t *cdb, uint8_t cdb_len, uint8_t *data,
     assert(cdb != NULL);
     assert(cdb_len != 0);
 
-    memset(&io_hdr, 0, sizeof(struct sg_io_hdr));
+    memset(&io_hdr, 0, sizeof(struct sg_io_v4));
```
tasleson: Good catch.  I'm thinking that we may avoid repeating this bug in the future if we did

```c
memset(&io_hdr, 0, sizeof(io_hdr));
```

That way the structure type itself can change and this memset will still be valid.

tasleson:
@cathay4t Thanks Gris

PR #312, opened 2017-10-02
Ontap: Remove m2crypto requirement

tasleson:


cathay4t:
@tasleson Could you make quick notes on why we need m2crypto in old and don't now? Thank you.

tasleson:
@cathay4t When we did the py3 conversion we removed the need to have m2crypto as we implemented RC4 in our own code base as it's quite trivial.

cathay4t:
Thanks. Merged.

PR #316, opened 2017-10-17
LocalDisk: Fix incorrect VPD query for 0x89 and 0xb1.

cathay4t:
Bug:    https://bugzilla.redhat.com/show_bug.cgi?id=1501367

Root cause:
```
    Incorrect data size for VPD 0x89(ATA information).
    The SPEC only allows 572(0x023c) bytes, we are requesting 0xffff
    bytes.
```

Fix:
```
    Change _sg_io_vpd() to set correct output data buffer size for
    page 0x89(ATA infomation) and 0xb1(Block Device Characteristics).
```

CI Kick

Also add newline to every warning message of lsmcli on stderr.

### DIFF
--- a/c_binding/libsg.c
+++ b/c_binding/libsg.c
@@ -121,6 +121,16 @@
  */
 #define _T10_SAT_ATA_PASS_THROUGH_PROTOCOL_NON_DATA     0x3
 
+/*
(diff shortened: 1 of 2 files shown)


tasleson:
@cathay4t Thanks Gris

PR #318, opened 2017-10-30
Enable RHEL 6 support and tag 1.6.1 release.

cathay4t:
CI Kick again!


tasleson:
@cathay4t The ppc64 system was missing the log output directory `/tmp/lsm_ci_log` which is weird as the code checks for it at start up.  So it appears that it was deleted after the service was started.  Additionally the code running on that node needs to be updated.

tasleson:
@cathay4t Tested EL6 build and `make distcheck`, looks good thanks!

PR #321, opened 2017-12-01
WIP: Rust binding of libstoragemgmt client API

cathay4t:
**Just for early review**

HowTo:
 * `sudo yum install libstoragemgmt -y`
 * `sudo systemctl start libstoragemgmt`
 * `cd rust_binding/ && cargo test`

Done:
 * All functions except LocalDisk functions.
 * All functions has unit tests.

TODO:
 * Documentation.
 * Include `cargo build` in `make`.
 * Include `cargo test` in `make check`.
 * Find a way to upload to http://crates.io/, It might need a standalone git project.

Notes:
 * No ASYNC actions any more. User should use thread instead.
 * No flags, no code is using flags yet. We may add new functions if new flags come in.


tasleson:
@cathay4t Why no async support in client lib?

tasleson:
@mulkieran FYI

rust_binding/src/error.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+}
+
+impl From<ErrorKind> for LsmError {
+    fn from(ek: ErrorKind) -> LsmError {
+        LsmError{kind: ek}
+    }
+}
+
+pub type Result<T> = result::Result<T, LsmError>;
```
tasleson: @cathay4t @mulkieran Is it a good idea to create the new type with the same name?  I know error chain does this, but...

cathay4t: Error chain cannot handle 42+ errors we needed:
https://github.com/rust-lang-nursery/error-chain/issues/117

cathay4t: Oh. Just save some typing. not a big deal.

cathay4t:
> @cathay4t Why no async support in client lib?
In current lsmcli, there are a lot lines handling this async, if I do that in rust, I had to guide user to the
same, even for simple task as `volume_create()`. And supporting async increase the API complexity.
Comparing to thread, the only async benefit is allowing user to query progress. I am willing to sacrifice that for clear code and simple API design.

cathay4t:
Closing to save unneeded CI works.
Will recreate PR if needed.

cathay4t:
rust binding created in separate repo:

 * crates.io: https://crates.io/crates/libstoragemgmt-rust
 * Github: https://github.com/libstorage/libstoragemgmt-rust
 * Doc: https://libstorage.github.io/libstoragemgmt-rust

PR #322, opened 2017-12-07
MegaRAID plugin: Fix cache information query.

cathay4t:
* MegaRAID always enable read cache. Old code treat that as read ahead
   cache which is not what we defined in API document.

 * Fix the write cache set function to handle 'Cache I/O'.

 * Inform user via volume_read_cache_policy_update(): read cache is
   always enabled, if they are looking for read ahead cache setting, they
   should use storcli/perccli.

### DIFF
--- a/plugin/megaraid/megaraid.py
+++ b/plugin/megaraid/megaraid.py
@@ -1067,9 +1067,6 @@ class MegaRAID(IPlugin):
         sys_all_output = self._storcli_exec(
             ["/%s" % vd_path.split('/')[1], "show", "all"])
 
-        write_cache_status = Volume.WRITE_CACHE_STATUS_WRITE_THROUGH
-        read_cache_status = Volume.READ_CACHE_STATUS_DISABLED
-
         ram_size = _mega_size_to_lsm(
             sys_all_output['HwCfg'].get('On Board Memory Size', '0 KB'))
         if ram_size > 0:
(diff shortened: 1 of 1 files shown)


plugin/megaraid/megaraid.py
```diff
@@ -1080,37 +1077,44 @@ def volume_cache_info(self, volume, flags=Client.FLAG_RSVD):
             if lsm_bat.status == Battery.STATUS_OK:
                 flag_battery_ok = True
 
+
```
tasleson: Added blank line that is not needed, pep8 should flag this

plugin/megaraid/megaraid.py
```diff
@@ -1181,8 +1200,9 @@ def volume_read_cache_policy_update(self, volume, rcp,
         storcli always enable read cache and no way to change it
         """
         raise LsmError(ErrorNumber.NO_SUPPORT,
-                       "LSI MegaRAID always enable read cache and refused to "
-                       "change that.")
+                       "MegaRAID always enable read cache and refused to "
+                       "change that. You can change read ahead cache "
+                       "setting via storcli/perccli")
```
tasleson: The plugin doesn't advertise that is supports the method `volume_read_cache_policy_update` so technically this method should never be called by a client, but I guess it's OK to have it.

cathay4t:
@tasleson All fixed. Thanks Tony.

tasleson:
@cathay4t Thanks Gris!

Issue #194, opened 2016-08-25
pyudev: Need to accomodate API changes


tasleson:
pyudev has taken the approach that they will deprecate and remove functionality with no consideration for library users .  We will either need to write a compatibility layer in libstoragemgmt to accommodate the differences and maintain it as pyudev continues to evolve or remove it.

~~I believe we should remove use of pyudev so that we can reduce our long term support burden.~~

tasleson:
The maintainer will adhere to versioning.  So when the 1.0.0 is released we will likely be broken, but should be fine after we make the needed changes.

cathay4t:
The only place using pyudev is hpsa plugin where we could simply replace it with some sysfs actions or even better lsm.LocalDisk functions.

joehandzik:
@cathay4t I'm personally indifferent, we had something that worked fine before but moved to pyudev because it was a cleaner implementation.

tasleson:
@joehandzik You are correct, sorry for the code churn.  Knowing what I know now, we would not have chosen to pick up the pyudev dependency.  I'm using pyudev for a couple of other things which are not this trivial, and I'm trying to figure out how I want to handle those cases too.

joehandzik:
@tasleson No worries! We can handle this however you guys like. I can see the value in using pyudev as a buffer between us and sysfs. If there's a reasonable way to keep pyudev, we should keep it.

cathay4t:
Closing this issue against PR #330.

PR #330, opened 2018-03-21
SmartArray plugin: Remove the pyudev codes.

cathay4t:
Removed the pyudev codes from hpsa plugin.

https://github.com/libstorage/libstoragemgmt/issues/194


tasleson:
Thanks Gris

PR #331, opened 2018-03-22
Fix compile warnnings on gcc 8

cathay4t:
Tested on Fedora rawhide `gcc-8.0.1-0.19.fc29.x86_64`

### DIFF
--- a/c_binding/libsg.c
+++ b/c_binding/libsg.c
@@ -367,8 +367,9 @@ static int _sg_io_open(char *err_msg, const char *disk_path, int *fd,
  * The 'sense_key' is the output pointer.
  * Return 0 if sense_key is _T10_SPC_SENSE_KEY_NO_SENSE or
  * _T10_SPC_SENSE_KEY_RECOVERED_ERROR, return -1 otherwise.
+ * sense_err_msg should be char[_LSM_ERR_MSG_LEN / 2]
  */
-static int _check_sense_data(char *err_msg, uint8_t *sense_data,
+static int _check_sense_data(char *sense_err_msg, uint8_t *sense_data,
                              uint8_t *sense_key);
 
 static int _extract_ata_sense_data(char *err_msg, uint8_t *sense_data,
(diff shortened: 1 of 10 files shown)


c_binding/libsg.c
```diff
@@ -485,14 +486,14 @@ int _sg_io_vpd(char *err_msg, int fd, uint8_t page_code, uint8_t *data)
...
-    char sense_err_msg[_LSM_ERR_MSG_LEN];
+    char sense_err_msg[_LSM_ERR_MSG_LEN / 2];
     ssize_t data_len = 0;
 
     assert(err_msg != NULL);
     assert(fd >= 0);
     assert(data != NULL);
 
-    memset(sense_err_msg, 0, _LSM_ERR_MSG_LEN);
+    memset(sense_err_msg, 0, sizeof(sense_err_msg)/sizeof(char));
```
tasleson: `memset` clears a number of bytes, the statement `sizeof(sense_err_msg)/sizeof(char)` gives you the count or number of items stored in the memory which is not what what we want.  In this case we happen to be OK as `sizeof(char)` happens to be 1, thus N/1 == N.  Please simply change all the `memset` calls to use `sizeof(thing)`.

tasleson:
@cathay4t Have you compiled this on RHEL6/7?

c_binding/libsg.c
cathay4t: I will fix it.

cathay4t:
The travis CI will check RHEL/Centos 7 using `make distcheck`.
Let me see whether I could add CI for RHEL/Centos 6.

cathay4t:
http://people.redhat.com/fge/lsm/libstoragemgmt-1.6.1-2.el7.src.rpm

cathay4t:
RHEL 6.9 'make distcheck' pass.

cathay4t:
New patch uploaded to enable travis CI for Centos 6 `make check`.

cathay4t:
GCC 8.1 released. PR tested against it.
Merged.

PR #333, opened 2018-06-12
Python plugin runner: don't print socket error to syslog.

cathay4t:
Treat `socket.error` as client went away if plugin is registered.

Fixing bug: https://bugzilla.redhat.com/show_bug.cgi?id=1582458


cathay4t:
Merged.

PR #336, opened 2018-06-25
lsmcli: Fix required to run on python3.7

cathay4t:
* Python 3.7 introduced new reserved keywords -- `async` which happened
   to be used in lsmcli. We change lsmcli to use `_async` instead.


cathay4t:
There is another new keywords they introduced -- `await`. FYI.

cathay4t:
Merged.

PR #338, opened 2018-08-16
Replace yajl with nlohmann/json

cathay4t:
**Don't commit yet, just want to try CI**

Notes:
* nlohmann/json require gcc 4.9+ while RHEL 7 is shipping 4.8.

TODO:
 * convscan
 * code self-review.


cathay4t:
PR is ready for review.

c_binding/lsm_plugin_ipc.cpp
```diff
@@ -1170,7 +1170,6 @@ static int ag_create(lsm_plugin_ptr p, Value & params, Value & response)
 
             lsm_access_group *ag = NULL;
             lsm_system *system = value_to_system(v_system);
-
```
tasleson: Seems like a random ws change?

c_binding/lsm_ipc.cpp
```diff
@@ -861,7 +538,8 @@ Value Ipc::readRequest(void)
 {
     int ec;
     std::string resp = t.msg_recv(ec);
-    return Payload::deserialize(resp);
+    Value rc = Payload::deserialize(resp);
+    return rc;
 }
```
tasleson: Any reason this needed to be changed?

tasleson:
@cathay4t Looks good to me.  However, I'm not sure at the moment how to best address the compiler support on RHEL7, I suppose utilizing Red Hat Developer Toolset would be problematic?

c_binding/lsm_plugin_ipc.cpp
cathay4t: Will revert that.

c_binding/lsm_ipc.cpp
cathay4t: No, will revert that.

cathay4t:
Let me try to upgrade libstoragemgmt in EPEL 6 to use developer toolset and find out the procedure of RHEL 7.

cathay4t:
The EPEL6 has no DTS support yet, but [Fedora guys could enable it per request][1].

For RHEL 7, I got the rpm spec ready and ABI tested, but requires some red tape.

I would suggest we keep using yajl in RHEL 6, 7(and other old GCC system) by 
maintaining a new branch `yajl`.


[1]: http://smoogespace.blogspot.com/2018/03/using-red-hat-developer-toolset-dts-in.html

tasleson:
@cathay4t I think it would be possible to refactor `lsm_ipc.hpp` to what is needed for nlohmann json and then provide two implementations of it by having one that is based on yajl and the other by nlohmann json and use conditional compilation to select which one.  Thus we wouldn't need a separate branch.  Does this make sense?

cathay4t:
Yes. I will try that. Thanks.

tasleson:
@cathay4t Built and tested with yajl and new json library.  Everything checked out.  Looks good to me, thanks!

Issue #340, opened 2018-08-30
PLUGIN_BUG(2): _hp_size_to_lsm(): Got unexpected HP size string 1.0


ramanduh:
Got a PLUGIN_BUG(2) error when running a volume-cache-info query:
```
# lsmcli -u 'hpsa://' volume-cache-info --vol 600508b1001c224ab970433265f36a55
PLUGIN_BUG(2): _hp_size_to_lsm(): Got unexpected HP size string 1.0
```
Here is my ssacli version:
```
# ssacli version

   SSACLI Version: 3.25.4.0 2017-12-04
   SOULAPI Version: 3.25.4.0 2017-12-04

```
This issue does not happen on older version of ssacli (v ssacli-3.10.3.0). In fact, it seems that the output is changed from ssacli 3.25. If we ran the following command `ssacli ctrl slot=1 show config detail`, we can see the following difference:
```
    Cache Status: OK
    Cache Ratio: 10% Read / 90% Write
    Drive Write Cache: Disabled
-   Total Cache Size: 1024 MB
-   Total Cache Memory Available: 816 MB
+   Total Cache Size: 1.0
+   Total Cache Memory Available: 0.8
    No-Battery Write Cache: Disabled
    SSD Caching RAID5 WriteBack Enabled: False
    SSD Caching Version: 1
```

cathay4t:
Closing as PR https://github.com/libstorage/libstoragemgmt/pull/341 fixed this problem.

PR #341, opened 2018-08-30
HPSA plugin: Support ssacli version 3.25+

cathay4t:
Since ssacli version 3.25, the 'Total Cache Size' is changed to number
based on GiB. Like below:

    -   Total Cache Size: 1024 MB
    -   Total Cache Memory Available: 816 MB
    +   Total Cache Size: 1.0
    +   Total Cache Memory Available: 0.8

This causes volume_raid_info() failed on unknown size string.

https://github.com/libstorage/libstoragemgmt/issues/340

Also fixed the running hpsa on python3 the dictview issue.

### DIFF
--- a/plugin/hpsa/hpsa.py
+++ b/plugin/hpsa/hpsa.py
@@ -286,7 +286,8 @@ _HP_VENDOR_RAID_LEVELS = ['1adm', '1+0adm']
 
 
 _LSM_RAID_TYPE_CONV = dict(
-    list(zip(list(_HP_RAID_LEVEL_CONV.values()), list(_HP_RAID_LEVEL_CONV.keys()))))
+    list(zip(list(_HP_RAID_LEVEL_CONV.values()),
+             list(_HP_RAID_LEVEL_CONV.keys()))))
 
 
 def _hp_raid_level_to_lsm(hp_ld):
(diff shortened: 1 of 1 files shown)


plugin/hpsa/hpsa.py
```diff
@@ -1183,19 +1185,26 @@ def volume_cache_info(self, volume, flags=Client.FLAG_RSVD):
...
 
-        if 'Total Cache Size' in ctrl_data and \
-           _hp_size_to_lsm(ctrl_data['Total Cache Size']) > 0 and \
-           ctrl_data['Cache Status'] == 'OK':
-            flag_ram_ok = True
+        if 'Total Cache Size' in ctrl_data:
+            cache_size_str = ctrl_data['Total Cache Size']
+            # Since ssacli version 3.25, cache size is a number based in GiB.
+            try:
+                int(cache_size_str)
```
ramanduh: Do you mean float here? In both case, we'll get a ValueError:
```
>>> int('1024 MB')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1024 MB'
>>> int('1.0')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1.0'
>>> 
```

cathay4t: Sorry, not testing on python2. Python3 treat `int(1.0)` as works.

cathay4t: Fixed.

cathay4t:
@ramanduh Can you confirm whether this PR fixed your issue? Thank you.

ramanduh:
@cathay4t Yes, it fixed the issue. Thank you

PR #352, opened 2018-11-29
Update json.hpp for GCC 4.8 support

sharath-t:
On RHEL 7.6 with GCC 'gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-36)' make command fails with "./json.hpp:127:10: error: #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"" error.
changed gcc version check condition from 40900 to 40800


tasleson:
@sharath-t Our plan was to use yajl for RHEL 6/7

PR #353, opened 2018-12-07
Misc. fixes

tasleson:
Corrections for a number of bugs that have been sitting around, more corrections to come.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 language: c
 env:
     matrix:
-        - OS_TYPE=fedora
+        - OS_TYPE=fedora:28
         - OS_TYPE=centos:7
         - OS_TYPE=centos:6
         - OS_TYPE=debian:testing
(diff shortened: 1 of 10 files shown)


tasleson:
@cathay4t If you don't have time to review this, no worries :-)

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: required
 language: c
 env:
     matrix:
-        - OS_TYPE=fedora
+        - OS_TYPE=fedora:28
```
cathay4t: The intention of  `OS_TYPE=fedora` was to use latest fedora stable release for testing without
changing this file every time Fedora tag a release.
Why do we need to stick on Fedora 28?

tasleson: F29 Only has python3 packages.  When we specify `--with-python3` for fedora we test a combination of py2 and py3 testing which fails as py2 packages aren't available.  I'll take that out of the test and change this back.

PR #359, opened 2019-01-30
Arcconf Plugin patch: Fix for 'Systems' function

sharath-t:
issue: column 'Read Cache Percentage' data displayed is -1
		fix: 'readCachePercentage' value is read from getconfigJSON and is displayed properly.

issue: column 'Info'  data displayed is "Controller Status"=[2] mode=[3]
                fix: corresponding display string for 'status' and 'mode' is set. added 'slot' information as well


sharath-t:
Please consider the private functions and constants newly added are as part of support for further PRs as i am splitting the earlier jumbo PR into multiple PRs

PR #362, opened 2019-01-30
Arcconf Plugin patch: Fix for 'disks' function

sharath-t:
issue: column 'Name' has a string prefix 'test'
		fix : replace the 'test' string with manufacturer name
		
issue: column 'vdp' and disk path is empty. no logic implemented
		fix: for the Drives having 'physicalDriveName', use the lsm native function 'LocalDisk.vpd83_get(disk_path)' to get the vdp info
		
issue: column 'Size' showing incorrect data
		fix: code was looking for 'physicalBlockSize' where this was different than 'blockSize'. changed the code to fetch 'blockSize'
		
issue: column 'Link Type' was displaying incorrect information
		fix: code was using the same native Disk constants to represent Disk Type and Link Type. implemented separate functions for Type and Link type

### DIFF
--- a/plugin/arcconf/arcconf.py
+++ b/plugin/arcconf/arcconf.py
@@ -484,14 +484,11 @@ class Arcconf(IPlugin):
     def _arcconf_disk_to_lsm_disk(arcconf_disk, sys_id, ctrl_num):
         disk_id = arcconf_disk['serialNumber'].strip()
 
-        disk_name = "%s" % (arcconf_disk['model'])
+        disk_name = "%s - %s" % (arcconf_disk['vendor'], arcconf_disk['model'])
         disk_type = _disk_type_of(arcconf_disk)
-        link_type = disk_type
+        link_type = _disk_link_type_of(arcconf_disk)
 
-        try:
-            blk_size = int(arcconf_disk['physicalBlockSize'])
-        except KeyError:
-            blk_size = int(arcconf_disk['blockSize'])
+        blk_size = int(arcconf_disk['blockSize'])
         blk_count = int(arcconf_disk['numOfUsableBlocks'])
 
         status = _disk_status_of(arcconf_disk)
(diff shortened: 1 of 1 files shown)


plugin/arcconf/arcconf.py
```diff
@@ -541,7 +539,7 @@ def disks(self, search_key=None, search_value=None,
                                     arcconf_disk, sys_id, cntrl_num))
 
         return search_property(rc_lsm_disks, search_key, search_value)
-
+
```
cathay4t: Please remove this white space.

plugin/arcconf/arcconf.py
```diff
@@ -484,14 +484,11 @@ def volumes(self, search_key=None, search_value=None,
...
+        disk_name = "%s - %s" % (arcconf_disk['vendor'], arcconf_disk['model'])
         disk_type = _disk_type_of(arcconf_disk)
-        link_type = disk_type
+        link_type = _disk_link_type_of(arcconf_disk)
 
-        try:
-            blk_size = int(arcconf_disk['physicalBlockSize'])
-        except KeyError:
-            blk_size = int(arcconf_disk['blockSize'])
+        blk_size = int(arcconf_disk['blockSize'])
```
cathay4t: @sharath-t Can you explain to me why `physicalBlockSize` is not used here?

plugin/arcconf/arcconf.py
sharath-t: Done - commit id - 6651667aa6ba35ac90226cfc320baa488fd94e90

plugin/arcconf/arcconf.py
sharath-t: Thanks for the catch.
i discussed internally and found that 'physicalBlockSize' was added to support 4kn drives. so i have reverted the code to earlier state.

tasleson:
Integrated into https://github.com/libstorage/libstoragemgmt/pull/386

PR #361, opened 2019-01-30
Arcconf Plugin patch: Fix for 'volumes' function

sharath-t:
issue: column 'size' is showing wrong value
		fix: changed num_of_blocks calculation equation from [(dataSpace*1024)/blockSize] to dataSpace(or chunk->0->numBlocks)
		
issue: pool ID is incorrect for the LDs which are part of the same array
		fix: read 'consumedArrayId' in LD->chunk->0 to get array id
		
issue: column 'id' is not unique. volumes created out of different controllers has repeated IDs
		fix: volume ID is prefixed with system ID [systemID:volumeID]

PR #364, opened 2019-01-30
Arcconf Plugin patch: Fix for 'volume-delete' function

sharath-t:
changes needed based on changes in volume() function (PR #361 ) to get proper volume ID.


sharath-t:
@cathay4t 'array' in arcconf is equivalent to 'pool'. thats the reason i changed it to 'vol_id'. and since it may be confused with libstoragemgmt.Volume.id , i have now changed it to 'ld_id' which is in relevance to arcconf terms.
+ i have made use of volume.plugin_data instead of volume.id

these are the changes in the new commit id - 5f608c2b9f7d47fa950a53aacdb934476b9ed643

tasleson:
Integrated into https://github.com/libstorage/libstoragemgmt/pull/386

PR #370, opened 2019-02-11
man: Remove SNAPSHOTS from optional --sys constraint

tasleson:
When a user lists snapshots it is required that they supply a
filesystem.  A filesystem is specific to a given system, thus
allowing the user to constrain the search to a specific system
with --sys <sys_id> is redundant.

CI redo


cathay4t:
LGTM. Thanks.

PR #375, opened 2019-02-26
megaraid: avoid INVALID_ARGUMENT error when disk WNN is NA

yosshy:
Old MegaRAID cards can show disks with WWN==NA. It causes INVALID_ARGUMENT like below:

$ export LSMCLI_URI=megaraid://
$ sudo -E lsmcli ld
INVALID_ARGUMENT(101): Incorrect format of VPD 0x83 NAA(3) string: 'na', expecting 32 or 16 lower case hex characters

This patch overwrites VPD 0x83 value with an empty string if it has 'NA'.


tasleson:
I'm thinking we should address this in the plugin instead of the CLI

PR #378, opened 2019-03-19
debian: fix duplicate files in packages

emmericp:
/usr/libexec/lsm.d/*.py was included in both python-libstoragemgmt and libstoragemgmt-tools, removing from libstoremgmt-tools as it depends on python-libstoragemgmt.


tasleson:
Thanks @emmericp

Issue #383, opened 2019-03-27
Unable to control some slot LEDs


hjmallon:
This was also posted to the mailing list but it seems a bit quiet there (sorry for x-post).

I am trying to use libstoragemgmt (via udisks) to control the ident and fault LEDs on a 12-disk storage device with an LSI 9400 SAS card. Everything works except the disk in slot “15” (slots 0, 2, 9 and 11 are not filled due to layout). For which the SAS address is not found in _ses_info_get_by_sas_addr. The invoked commands seem to return a list of addresses for every slot (even the empty ones) *except* slot 15. Does anyone know why this might happen and/or what can be done to fix it?

tasleson:
Just to clarify, the the storage device has 16 slots, but only 12 are filled and the last slot, 15 is missing?  We may need to build a version with more debug unless @cathay4t has some insight into what might be going on.

tasleson:
@hjmallon Also sorry for not responding on mailing list and thanks for posting here too!

hjmallon:
Yeah we have 4 backplanes, each 4 slots, but some of them are not filled with disks. All the slot LEDs can be accessed except number 15.

tasleson:
@hjmallon I think it would be helpful to know what the contents of `/sys/class/bsg/*` look like on your system and the value of each `/sys/class/bsg/<name>/device/type`.

hjmallon:
```
# ls -1 /sys/class/bsg/
5:0:0:0
8:0:0:0
8:0:1:0
8:0:10:0
8:0:11:0
8:0:12:0
8:0:2:0
8:0:3:0
8:0:4:0
8:0:5:0
8:0:6:0
8:0:7:0
8:0:8:0
8:0:9:0
end_device-8:0
end_device-8:1
end_device-8:10
end_device-8:11
end_device-8:12
end_device-8:2
end_device-8:3
end_device-8:4
end_device-8:5
end_device-8:6
end_device-8:7
end_device-8:8
end_device-8:9
sas_host8
```

```
# for i in /sys/class/bsg/*/device/type; do echo "$i"; cat "$i"; done
/sys/class/bsg/5:0:0:0/device/type
0
/sys/class/bsg/8:0:0:0/device/type
0
/sys/class/bsg/8:0:10:0/device/type
0
/sys/class/bsg/8:0:1:0/device/type
0
/sys/class/bsg/8:0:11:0/device/type
0
/sys/class/bsg/8:0:12:0/device/type
13
[...]

PR #384, opened 2019-03-27
Fix local-disk-*-led-* on final slot in SES list

hjmallon:
The PAGE LENGTH in the SES-3 spec is labelled 'n - 3' but that doesn't mean we should do PAGE LENGTH + 3 to get the buffer size as the 0 byte is being counted.

Fixes: https://github.com/libstorage/libstoragemgmt/issues/383

CI Kick


tasleson:
Thanks @hjmallon

Issue #374, opened 2019-02-23
Change System.id of arcconf to use Controller Serial Number


cathay4t:
Currently, the `System.id` of arcconf is string like `1`.

This is not a unified id. Use `Controller Serial Number` or `Controller World Wide Name` could be better.

tasleson:
@sharath-t Is this something you could help out with?

sharath-t:
sure, i'll try to take this up in next couple of days.

PR #388, opened 2019-04-10
Change System.id of arcconf to use Controller Serial Number

sharath-t:
This PR is for the issue https://github.com/libstorage/libstoragemgmt/issues/374

This PR has changes in 3 public functions - 'systems()', 'pools()' and 'volumes()' and 2 private functions '_arcconf_array_to_lsm_pool()' and '_arcconf_ld_to_lsm_vol()'

* since 'System ID' is displayed in list type of 'volumes' and 'pools', changes are made in those functions to have controller serial number as sys_id
* since 'arcconf getconfigJSON' is dependent on the controller number, additional variable 'arcconf_ctrl_num' is introduced to have controller number in 'pools()' and 'volumes()'
* APIs dependent on pools and volumes need to read controller numbers as well. hence the 'plugin_data' variable in pools and volumes will have 'ctrl_num' instead of 'sys_id'

CI Redo

### DIFF
--- a/plugin/arcconf/arcconf.py
+++ b/plugin/arcconf/arcconf.py
@@ -379,7 +379,7 @@ class Arcconf(IPlugin):
         for cntrl in range(total_cntrl):
             ctrl_info = getconfig_cntrls_info[cntrl]['Controller']
             ctrl_name = "%s %s" % (ctrl_info['deviceVendor'], ctrl_info['deviceName'])
-            sys_id = str(cntrl + 1)
+            sys_id = ctrl_info['serialNumber']
             status = ctrl_info['controllerStatus']
             if status == CONTROLLER_STATUS_WORKING:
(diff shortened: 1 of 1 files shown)


sharath-t:
@cathay4t i have created a single PR for changes in multiple public functions as it is related to the same issue. kindly accept.

sharath-t:
sysid_changes

attached screenshot on how the change looks

tasleson:
@sharath-t I think the system id looks good, but it appears the volume id shares the same value as the pool id?

sharath-t:
> @sharath-t I think the system id looks good, but it appears the volume id shares the same value as the pool id?

@tasleson yes, since the complete pool is carved to form a volume both has same ID here. it's the direct value of array ID and logical ID from arcconf.

tasleson:
@sharath-t Thanks

PR #390, opened 2019-04-11
System id update in disks function

sharath-t:
As part of https://github.com/libstorage/libstoragemgmt/issues/374 . system id in disks was left out in previous PR.


sharath-t:
@tasleson i just noticed the system id update in disks display was left out. please accept the PR

tasleson:
@sharath-t  Good catch, thanks!

PR #405, opened 2019-12-03
Moved from strerror_r to strerror_l to compile on Alpine

### LINKED ISSUE #401: libstoragemgmt doesn't compile on Alpine Linux (musl)
Mainly due to only having POSIX stderr_r which I believe is deprecated in favor of stderr_l. I have created a quick and dirty patch to switch to stderr_l and have compiled successfully.

PhoenixMage:
Fixes #401 

CI Test

### DIFF
--- a/c_binding/libses.c
+++ b/c_binding/libses.c
@@ -439,6 +439,7 @@ static int _ses_bsg_paths_get(char *err_msg, char ***bsg_paths,
         *bsg_paths = NULL;
         *bsg_count = 0;
     }
+    freelocale(locale);
     return rc;
 }
 
--- a/c_binding/libses.c
+++ b/c_binding/libses.c
@@ -16,9 +16,6 @@
  * Author: Gris Ge <fge@redhat.com>
  */
 
-/* For strerror_r() */
-#define _GNU_SOURCE
-
 #include "libsg.h"
 #include "libses.h"
 #include "utils.h"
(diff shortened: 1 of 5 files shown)


PhoenixMage:
Not sure why the last commit failed? It is successfully passing the test-suite on my local ubuntu box. The last commit should have fixed the issue its failing on?

tasleson:
@PhoenixMage I'm wondering if we can do something like this instead, replace calls of `strerror_r` with this, to simply things like remembering to free locale.  Only question is if it's valid to reference the string after calling `freelocale`.  Some simple testing seems to indicate yes, but that's not conclusive.

```C
const char *errno_text(int error_number)
{
    const char *non_mem = "Unable to translate errno to text, newlocale fail";
    locale_t const locale = newlocale(LC_MESSAGES_MASK, "", (locale_t)0 );
    if (!locale) {
        return non_mem;
    }
    char *s = strerror_l(error_number, locale);
    freelocale(locale);
    return s;
}
```

PhoenixMage:
@tasleson I am no good with C in general but reading the documentation for freelocale() it appears the behaviour after freeing is undefined so probably wouldnt recommend it.

https://pubs.opengroup.org/onlinepubs/9699919799/functions/freelocale.html

tasleson:
> reading the documentation for freelocale() it appears the behaviour after freeing is undefined so probably wouldnt recommend it.

@PhoenixMage  Well the documentation discusses the locale data, but the string that has been created should be OK.  However, we can be very sure if we copy string to our own local buffer before freeing the locale.  I'll put something together tomorrow.  Thanks for the PR's and compiling on alpine!

PR #428, opened 2020-08-10
tester.c: Update 'check' usage

tasleson:
The unit test library check made a change which emits warnings
when 'fail_unless' is passed a message.  For more information
see:

https://github.com/libcheck/check/pull/298/commits/82540c5428d3818b64d6a8aefb601e722520651f

As we compile with warnings as errors, we fail to compile when
using check >= 0.15.0.

Remove use of deprecated 'fail_unless'.


tasleson:
Added a couple of commits which re-format all the C code in the repo, not just that in `c_binding` directory.  Not sure why I did this originally, but all the code benefits from consistent style.

Issue #420, opened 2020-05-12
lsmcli -u hpsa:// ld PLUGIN_BUG(2): Expected key missing from SmartArray ssacli output:'Serial Number'


cstoyanov-scality:
* libstoragemgmt version: 1.7.3 and 1.8.0
* ssacli version: 3.30, 3.40 and 4.17

### Description

When the Serial Number of the physical disk cannot be retrieved, lsmcli fails with a PLUGIN_BUG error. For example:

```
$ lsmcli -u hpsa:// ld
PLUGIN_BUG(2): Expected key missing from SmartArray ssacli output:'Serial Number', please try to upgrade ssacli tool
```

Depending on the way the disk is broken, the ssacli report can contains or not some info. In the hpsa plugin of lsm, some information are expected to exist (if not, it raises with a KeyError exception).

the ssacli output which cannot be parsed by lsm is:

```
Smart Array P420i in Slot 1
   Bus Interface: PCI
   Slot: 1

[...]

   Physical Drives
      physicaldrive 2I:1:1 (port 2I:box 1:bay 1, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:2 (port 2I:box 1:bay 2, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:3 (port 2I:box 1:bay 3, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:4 (port 2I:box 1:bay 4, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:5 (port 2I:box 1:bay 5, SATA HDD, 4 GB, Failed)
      physicaldrive 2I:1:6 (port 2I:box 1:bay 6, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:7 (port 2I:box 1:bay 7, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:8 (port 2I:box 1:bay 8, SATA HDD, 0 GB, Failed)
      physicaldrive 2I:1:9 (port 2I:box 1:bay 9, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:10 (port 2I:box 1:bay 10, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:11 (port 2I:box 1:bay 11, SATA HDD, 4 TB, OK)
      physicaldrive 2I:1:12 (port 2I:box 1:bay 12, SATA HDD, 4 TB, OK)
[...]

PR #435, opened 2020-10-27
Smart array: Handle errored disk

tasleson:
Resolves: https://github.com/libstorage/libstoragemgmt/issues/420

Note: This change is an adaptation of the one provided by
Christophe Stoyanov.  Specifically, when we cannot retrieve all
the needed values we select default values of the proper type
and domain eg.

id, disk_name -> string
blk_size, blk_count -> unsigned integer
rpm -> Disk.RPM_UNKNOWN


tasleson:
@cstoyanov-scality Please review and test if you can.

Issue #68, opened 2015-10-30
lsm_disk_record_alloc() does not have plugin_data.


cathay4t:
Since API has released, we can add it in the way of `lsm_system_mode_set()`.

PR #459, opened 2021-03-22
c_binding: Add disk record plugin_data

tasleson:
This was missed in the original release.  Adding an additional
function to allow it to be utilized.

Resolves: https://github.com/libstorage/libstoragemgmt/issues/68

### DIFF
--- a/c_binding/lsm_datatypes.hpp
+++ b/c_binding/lsm_datatypes.hpp
@@ -276,6 +276,7 @@ struct LSM_DLL_LOCAL _lsm_disk {
     const char *location;
     int32_t rpm;
     lsm_disk_link_type link_type;
+    char *plugin_data; /**< Reserved for the plugin to use, added in 1.9 */
 };
 
 #define LSM_HASH_MAGIC   0xAA7A0011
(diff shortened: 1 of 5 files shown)


tasleson:
@trgill @tbzatek  If you folks have time I would appreciate a code review for this.  Specifically ensuring that I'm not breaking any backwards code compatibility while introducing this additional functionality, thanks!

c_binding/lsm_datatypes.hpp
```diff
@@ -276,6 +276,7 @@ struct LSM_DLL_LOCAL _lsm_disk {
     const char *location;
     int32_t rpm;
     lsm_disk_link_type link_type;
+    char *plugin_data; /**< Reserved for the plugin to use, added in 1.9 */
```
tbzatek, line 279: Growing the size of a presumably public `struct`. I think this should be safe in case new struct members are added at the end. No objections with the added functions.

tasleson: Thanks for taking a look.  The structure is also opaque, so code outside the library has no knowledge of internals.

PR #469, opened 2021-06-02
Correct endianess issues for big endian

tasleson:


tasleson:
The change for link speed is actually incorrect.  The problem is in the linux kernel itself.

PR #471, opened 2021-06-28
Use latest fedora again

tasleson:
Let's see if the issue we ran into before with F34 is still present.


tasleson:
Latest circle CI isn't working, closing.

PR #475, opened 2021-07-11
Debian-like OS fixes for python3 builds

nedvna:
Python2.x is deprecated, thus it's a good idea to finally move to Python3.

I suggest these 2 initial fixes that still let you build 'as is' (i.e. don't break Python2 compatibility for Debian-like OS) but let you build for Python3, if you make some additional changes.  
I'd also like to suggest changes to build on Python3 by default but that should be in another PR, I think.


tasleson:
Thanks for this PR.  I believe the reason we haven't moved to py3 was that some of the dependencies aren't yet available for py3 for debian/ubuntu, but I need to re-investigate.

tasleson (approved):

PR #494, opened 2022-01-28
Fix cmd and test

tasleson:
Correcting and test for `argparse` behavior change.


lkx007:
您好，您的来信已收到，我将会尽快给您回复，谢谢~~

PR #511, opened 2022-09-28
Add a udev rule entry for ALUA state changes

brianatpurestorage:
The Linux kernel has support for this event and will emit an event if a target returns this unit attention.

Add it commented out to match the others in the file.


e-milne:
Looks fine to me, if for some reason userspace wanted to handle an event.
The kernel would automatically update the ALUA state via RTPG *if* scsi_dh_alua
is attached, it doesn't have to be, though.

PR #528, opened 2023-04-14
Correction for fips error

tasleson:
When running on a fips enabled system we encounter the following error:

```
PLUGIN_BUG(2): [digital envelope routines] unsupported Data: Traceback (most recent call last): ...
  File "/usr/lib64/python3.9/site-packages/lsm/_common.py", line 348, in md5
    h = hashlib.md5()
ValueError: [digital envelope routines] unsupported
```

Utilize the `usedforsecurity=False`parameter to md5 to indicate that our use is not for security related purposes.


lkx007:
您好，您的来信已收到，我将会尽快给您回复，谢谢~~

Issue #135, opened 2016-05-02
How about loosen restriction of VPD83 id to support NVMe disks.


cathay4t:
@joehandzik @tasleson 

Currently, we are defining the VPD83 ID as VPD83 NAA ID which is not supported by NVMe disks yet.

The NVMe disks only expose SCSI VPD83 string ID which translate from NVMe serial:
```
[fge@megadeth ~]$ sudo sg_vpd --page=0x83 /dev/nvme0n1
Device Identification VPD page:
  Addressed logical unit:
    designator type: SCSI name string,  code set: UTF-8
      SCSI name string:
      8086INTEL SSDPEDMD400G4                     1000CVFT5194002T400BGN  
```

How about we make an exception in `lsm_local_disk_vpd83_get()` and `lsm_local_disk_vpd83_search()`
to allow VPD83 sting ID as fallback?

joehandzik:
I think that's fine, Gris. I saw some similar trickery in Ceph's spdk usage.

Joe

On May 2, 2016, at 7:21 AM, Gris Ge <notifications@github.com<mailto:notifications@github.com>> wrote:


@joehandzik<https://github.com/joehandzik> @tasleson<https://github.com/tasleson>

Currently, we are defining the VPD83 ID as VPD83 NAA ID which is not supported by NVMe disks yet.

The NVMe disks only expose SCSI VPD83 string ID which translate from NVMe serial:

[fge@megadeth ~]$ sudo sg_vpd --page=0x83 /dev/nvme0n1
Device Identification VPD page:
  Addressed logical unit:
    designator type: SCSI name string,  code set: UTF-8
      SCSI name string:
      8086INTEL SSDPEDMD400G4                     1000CVFT5194002T400BGN


How about we make an exception in lsm_local_disk_vpd83_get() and lsm_local_disk_vpd83_search()
to allow VPD83 sting ID as fallback?

-

cathay4t:
Another update on NVMe, with NVMe 1.2+(not sure what does this mean yet), linux kernel expose VPD83 EUI64 ID for NVMe disks.

joehandzik:
I guess we have three options (as far as I see):

1. Stick with the VPD83 string ID for all NVMe devices (I'm assuming they still expose it even with NVMe 1.2+)
2. Make the EUI64 ID the desired ID for NVMe devices, and fall back to the VPD83 string ID when it's not present (kind of like how we handle the the sas address sysfs attribute versus the value from the page inquiry)
[...]

PR #557, opened 2026-05-27
Improve unique identifier support to include NVMe devices

tasleson:
https://github.com/libstorage/libstoragemgmt/issues/135 discussed the lack of NVMe support.  This pull request addresses this need.

Additionally, a small update to hpsa plugin was done, so that we could have this information available when it's not available once the storage devices are part of a RAID group.

## Summary by CodeRabbit

* **New Features**
  * Enhanced NVMe device support with improved serial number and VPD 0x83 identification, including NVMe-specific probing and conversion.
  * Added SPC-5 VPD designation descriptor type support for EUI-64.
* **Bug Fixes**
  * Improved VPD 0x83 handling: broadened accepted hex lengths (16/24/32) and reduced over-restrictive validation so provided values are preserved more consistently across bindings/plugins.
  * Improved SCSI/NVMe VPD 0x83 resolution (including EUI-64 fallback) and corrected NVMe RPM/link type classification.
  * Improved HPSa disk identification when Disk Name is missing.
* **Documentation**
  * Updated installation guide bootstrapping guidance and expanded VPD 0x83 API documentation.

CI Kick

### DIFF
--- a/c_binding/lsm_local_disk.cpp
+++ b/c_binding/lsm_local_disk.cpp
@@ -8,6 +8,7 @@
  */
 
 #include <assert.h>
+#include <ctype.h>
 #include <dirent.h>
 #include <endian.h>
 #include <errno.h>
@@ -61,7 +62,14 @@
 
 #define _SYSFS_BLK_PATH_FORMAT      "/sys/block/%s"
 #define _MAX_SYSFS_BLK_PATH_STR_LEN (128 + _MAX_SD_NAME_STR_LEN)
-#define _SYSFS_SAS_ADDR_LEN         (_SG_T10_SPL_SAS_ADDR_LEN + 2)
+
+#define _SYSFS_NGUID_PATH_FORMAT            "/sys/block/%s/nguid"
+#define _MAX_SYSFS_NGUID_PATH_STR_LEN       (128 + _MAX_SD_NAME_STR_LEN)
+#define _SYSFS_EUI_PATH_FORMAT              "/sys/block/%s/eui"
+#define _MAX_SYSFS_EUI_PATH_STR_LEN         (128 + _MAX_SD_NAME_STR_LEN)
+#define _SYSFS_NVME_SERIAL_PATH_FORMAT      "/sys/block/%s/device/serial"
+#define _MAX_SYSFS_NVME_SERIAL_PATH_STR_LEN (128 + _MAX_SD_NAME_STR_LEN)
+#define _SYSFS_SAS_ADDR_LEN                 (_SG_T10_SPL_SAS_ADDR_LEN + 2)
 /* ^ Only Linux sysfs entry /sys/block/sdx/device/sas_address which
  *   format is '0x<hex_addr>\0'
  */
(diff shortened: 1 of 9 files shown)


plugin/hpsa_plugin/hpsa.py
```diff
@@ -760,7 +760,8 @@ def _hp_disk_to_lsm_disk(hp_disk,
         if disk_path:
             vpd83 = LocalDisk.vpd83_get(disk_path)
         else:
-            vpd83 = ''
+            unique_id = hp_disk.get('Drive Unique ID', '')
+            vpd83 = unique_id.lower()
```
trgill, line 765: Drive Unique ID is passed to the Disk constructor unvalidated.

The old code set vpd83 = '' in this branch, now, if Drive Unique ID is present but isn't exactly 16/24/32 hex chars — embedded spaces, a separator, a prefix, an odd length — Disk.__init__ raises INVALID_ARGUMENT and that aborts the whole disk enumeration, not just that one disk's vpd83. 

That's a behavioral change for any controller/firmware whose field isn't a clean hex string. 

Maybe strip whitespace, and ideally only assign it if it matches _disk_regex_vpd83, falling back to '' otherwise so a malformed value degrades gracefully instead of throwing.

c_binding/lsm_local_disk.cpp
```diff
@@ -678,6 +828,13 @@ int lsm_local_disk_rpm_get(const char *disk_path, int32_t *rpm,
 
     _lsm_err_msg_clear(err_msg);
 
+    *lsm_err = NULL;
+
+    if (strncmp(disk_path, "/dev/nvme", strlen("/dev/nvme")) == 0) {
+        *rpm = LSM_DISK_RPM_NON_ROTATING_MEDIUM;
+        goto out;
+    }
```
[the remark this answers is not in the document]
tasleson: Done

plugin/hpsa_plugin/hpsa.py
tasleson: @trgill I believe this is addressed

trgill:
looks good!

tasleson:
> looks good!

I'm actually thinking that having all this validation is the wrong approach.

The library is a transport layer between storage backends (plugins, sysfs,
ssacli) and consumers of the API. The sources are authoritative, the kernel wrote that sysfs file,
ssacli reported that Drive Unique ID, the SCSI VPD page has binary data that gets hex-encoded by
construction.  The library isn't going to correct it.  By placing all these restrictions what we actually got will be lost as it won't be given to the user.  It would probably be better to give the user what we got, so that they can report the problem correctly.

I'm going to think about this some more.

tasleson:
@trgill  I re-worked this quite a bit, to remove the library from the business of enforcing the id format.  Hopefully the commit messages covey the intent correctly.

tasleson:
@trgill When you get a chance, please take a peek at this latest version

trgill:
this is even better. lgtm

PR #560, opened 2026-06-17
configure.ac: make ledmon optional, fix systemd variable name mismatch

brianatpurestorage:
## Summary

Two fixes to `configure.ac` to improve portability across distributions:

- **Make ledmon optional**: `libled` (ledmon) was a hard build requirement via `PKG_CHECK_MODULES`, preventing builds on distributions that do not package libled as a development library (e.g. Debian, Ubuntu). This adds `--with-ledmon` / `--without-ledmon` configure flags. When ledmon is absent or disabled, all LED API functions return `LSM_ERR_NO_SUPPORT` cleanly, matching the behaviour of other optional features in the library.

- **Fix systemd `AC_ARG_WITH` variable name mismatch**: `AC_ARG_WITH` converts hyphens to underscores in shell variable names, so `--with-systemd-systemunitdir` sets `$with_systemd_systemunitdir`. The three systemd `AC_ARG_WITH` blocks were missing `ACTION-IF-GIVEN` clauses, so the intended variables were never assigned and `HAVE_SYSTEMD` was always false even when systemd was present. Also adds `pkg-config` fallback defaults for environments where `pkg-config systemd` returns an empty path.

## Test plan

- [ ] Build with ledmon present (`--with-ledmon`): LED API functions work as before
- [ ] Build without ledmon (`--without-ledmon`): library builds cleanly; LED API functions return `LSM_ERR_NO_SUPPORT`
- [ ] Build on Debian bookworm (no libled-dev): builds successfully with `--without-ledmon`
- [ ] Verify systemd service/sysusers/tmpfiles files install correctly when libsystemd-dev is present

## Summary by CodeRabbit

* **New Features**
  * Added a `--without-ledmon` build option to optionally disable LED monitoring support during installation.

* **Improvements**
  * LED operations now gracefully report unavailability when LED monitoring is disabled.
  * LED status/behavior is adjusted for non-SCSI controllers.
  * Enhanced systemd installation path configuration with improved fallback handling.

CI-KICK

### DIFF
--- a/c_binding/Makefile.am
+++ b/c_binding/Makefile.am
@@ -3,12 +3,19 @@ SUBDIRS = include
 AM_CPPFLAGS = 	-I$(top_srcdir)/c_binding/include \
 				-I$(top_builddir)/c_binding/include \
 				$(LIBGLIB_CFLAGS) \
-				$(LIBUDEV_CFLAGS) \
-				$(LIBLED_CFLAGS)
+				$(LIBUDEV_CFLAGS)
+
+if WITH_LEDMON
+AM_CPPFLAGS += $(LIBLED_CFLAGS)
+endif
 
 lib_LTLIBRARIES = libstoragemgmt.la
 
-libstoragemgmt_la_LIBADD=$(LIBXML_LIBS) $(LIBGLIB_LIBS) $(LIBUDEV_LIBS) $(LIBLED_LIBS)
+libstoragemgmt_la_LIBADD=$(LIBXML_LIBS) $(LIBGLIB_LIBS) $(LIBUDEV_LIBS)
+
+if WITH_LEDMON
+libstoragemgmt_la_LIBADD += $(LIBLED_LIBS)
+endif
 
 libstoragemgmt_la_LDFLAGS= -version-info $(LIBSM_LIBTOOL_VERSION)
 libstoragemgmt_la_SOURCES= \
--- a/c_binding/lsm_datatypes.hpp
+++ b/c_binding/lsm_datatypes.hpp
@@ -16,7 +16,9 @@
 #include "libstoragemgmt/libstoragemgmt_plug_interface.h"
 #include "lsm_ipc.hpp"
 #include <glib.h>
+#ifdef HAVE_LEDMON
 #include <led/libled.h>
+#endif
 
 #ifdef __cplusplus
 extern "C" {
@@ -313,21 +315,33 @@ struct LSM_DLL_LOCAL _lsm_battery {
 #define LSM_IS_LED_HANDLE(obj) MAGIC_CHECK(obj, LSM_LED_HANDLE_MAGIC)
 struct LSM_DLL_LOCAL _lsm_led_handle {
     uint32_t magic;
+#ifdef HAVE_LEDMON
     struct led_ctx *ctx;
+#else
+    void *ctx;
+#endif
 };
 
 #define LSM_LED_SLOT_MAGIC         0xAA7A0016
 #define LSM_IS_LED_SLOT_ENTRY(obj) MAGIC_CHECK(obj, LSM_LED_SLOT_MAGIC)
 struct LSM_DLL_LOCAL _lsm_led_slot {
     uint32_t magic;
+#ifdef HAVE_LEDMON
     struct led_slot_list_entry *slot;
+#else
+    void *slot;
+#endif
 };
 
 #define LSM_LED_SLOT_ITR_MAGIC   0xAA7A0015
 #define LSM_IS_LED_SLOT_ITR(obj) MAGIC_CHECK(obj, LSM_LED_SLOT_ITR_MAGIC)
 struct LSM_DLL_LOCAL _lsm_led_slot_itr {
     uint32_t magic;
+#ifdef HAVE_LEDMON
     struct led_slot_list *sl;
+#else
+    void *sl;
+#endif
     struct _lsm_led_slot entry;
 };
 
--- a/c_binding/lsm_local_disk.cpp
+++ b/c_binding/lsm_local_disk.cpp
@@ -12,7 +12,9 @@
 #include <endian.h>
 #include <errno.h>
 #include <fcntl.h>
+#ifdef HAVE_LEDMON
 #include <led/libled.h>
+#endif
 #include <libudev.h>
 #include <limits.h>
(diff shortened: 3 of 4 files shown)


tasleson:
@brianatpurestorage Thank you for this PR.  I try to get to reviewing it, likely on 6/22.

tasleson:
@brianatpurestorage 

Thanks for working on this, making ledmon optional is the right call for portability, and the systemd $withval fix is correct and needed. A few issues I spotted:

 ### 1. Hardcoded systemd fallbacks force HAVE_SYSTEMD=true on non-systemd systems

  The fallback logic:

  `test -n "$with_systemdsystemunitdir" || with_systemdsystemunitdir=/lib/systemd/system`

  means these variables are never empty. Since HAVE_SYSTEMD at line 387 checks `test -n "$with_systemdsystemunitdir" ...`, it will now always evaluate true, even on Alpine, minimal containers, or any system without systemd. This installs systemd unit/sysusers/tmpfiles to hardcoded paths where they don't belong.

 Previously, empty pkg-config output → empty variable → HAVE_SYSTEMD false. That was correct.

  Suggested fix, only fall back when systemd is actually present:

```bash
   [with_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd 2>/dev/null)
   if test -z "$with_systemdsystemunitdir" && $PKG_CONFIG --exists systemd 2>/dev/null; then
       with_systemdsystemunitdir=/lib/systemd/system
   fi]
```

  Same pattern for the sysusersdir and tmpfilesdir blocks.

 ### 2. Stub functions don't null-initialize output parameters

  The lsm_led_handle_get stub:

```c
  int lsm_led_handle_get(lsm_led_handle **handle, lsm_flag flags) {
      (void)handle; (void)flags;
      return LSM_ERR_NO_SUPPORT;
  }
```

  If a caller doesn't pre-initialize handle and later passes it to lsm_led_handle_free, MAGIC_CHECK de-references a garbage pointer, then interesting things happen. Same issue in `lsm_led_slot_iterator_get` for `*slot_itr` and `*lsm_err.`

 (The real implementation has similar gaps on some error paths, so this is a pre-existing pattern, but the stubs are a good place to do it right.)

  Suggested fix:

```c
  int lsm_led_handle_get(lsm_led_handle **handle, lsm_flag flags) {
      (void)flags;
      if (handle) *handle = NULL;
      return LSM_ERR_NO_SUPPORT;
  }
```

```c
  int lsm_led_slot_iterator_get(lsm_led_handle *handle,
                                lsm_led_slot_itr **slot_itr, lsm_error **lsm_err,
                                lsm_flag flags) {
      (void)handle; (void)flags;
      if (slot_itr) *slot_itr = NULL;
      if (lsm_err) *lsm_err = NULL;
      return LSM_ERR_NO_SUPPORT;
  }
 ```

### 3. Inconsistent `LSM_DLL_EXPORT` on stub functions

  Some stubs have LSM_DLL_EXPORT (lsm_led_slot_iterator_free, lsm_led_slot_id, etc.) while others don't
  (lsm_led_handle_get, lsm_led_slot_status_set, etc.). The project doesn't use -fvisibility=hidden so this doesn't cause missing symbols today, but it's inconsistent and would break if visibility settings ever tighten.

  Suggested fix, match the header declarations, e.g.:

```c
  int LSM_DLL_EXPORT lsm_led_handle_get(lsm_led_handle **handle, lsm_flag flags) { ... }
  void LSM_DLL_EXPORT lsm_led_handle_free(lsm_led_handle *handle) { ... }
  int LSM_DLL_EXPORT lsm_led_slot_iterator_get(...) { ... }
  lsm_led_slot LSM_DLL_EXPORT *lsm_led_slot_next(...) { ... }
  uint32_t LSM_DLL_EXPORT lsm_led_slot_status_get(lsm_led_slot *slot) { ... }
  int LSM_DLL_EXPORT lsm_led_slot_status_set(...) { ... }
```

 ### 4. Makefile.am not conditionalized for LIBLED

  `c_binding/Makefile.am` unconditionally references `$(LIBLED_CFLAGS)` and `$(LIBLED_LIBS)`. When ledmon is disabled these expand to empty so the build works, but it's fragile and potentially confusing. The AM_CONDITIONAL([WITH_LEDMON], ...) is already defined in configure.ac but never used.

  Suggested fix:

```bash
  AM_CPPFLAGS = -I$(top_srcdir)/c_binding/include \
                -I$(top_builddir)/c_binding/include \
                $(LIBGLIB_CFLAGS) \
                $(LIBUDEV_CFLAGS)

  if WITH_LEDMON
  AM_CPPFLAGS += $(LIBLED_CFLAGS)
  endif

  libstoragemgmt_la_LIBADD = $(LIBXML_LIBS) $(LIBGLIB_LIBS) $(LIBUDEV_LIBS)

  if WITH_LEDMON
  libstoragemgmt_la_LIBADD += $(LIBLED_LIBS)
  endif
  ```

brianatpurestorage:
Thanks for the detailed review. Updated to address all four points:

1. **Systemd fallbacks now gated on systemd presence** — fallback paths are only used when `pkg-config --exists systemd` succeeds, so non-systemd systems (Alpine, minimal containers) correctly get `HAVE_SYSTEMD=false`.

2. **Stub output parameters null-initialized** — `lsm_led_handle_get` now sets `*handle = NULL` and `lsm_led_slot_iterator_get` sets `*slot_itr = NULL` and `*lsm_err = NULL` before returning `LSM_ERR_NO_SUPPORT`. Same fix applied to `lsm_led_slot_status_set` for `*lsm_err`.

3. **LSM_DLL_EXPORT added consistently to all stubs** — matches the header declarations for `lsm_led_handle_get`, `lsm_led_handle_free`, `lsm_led_slot_iterator_get`, `lsm_led_slot_next`, `lsm_led_slot_status_get`, and `lsm_led_slot_status_set`.

4. **Makefile.am conditionalized with `if WITH_LEDMON`** — `$(LIBLED_CFLAGS)` and `$(LIBLED_LIBS)` are now only included when ledmon is enabled, using the `AM_CONDITIONAL` already defined in configure.ac.

tasleson:
> Thanks for the detailed review. Updated to address all four points:

Hmmm, I'm not seeing any updates to the commits, did you forget to push your changes?

tasleson:
LGTM, thank you, merged!

PR #562, opened 2026-06-24
debian: update packaging for 1.10.3 re-introduction

brianatpurestorage:
## Summary

Updates the Debian packaging to build and install correctly on modern Debian/Ubuntu systems. Verified working on Debian 12 (bookworm), Debian 13 (trixie), and Ubuntu 24.04 (arm64).

The 2016 packaging predates several upstream changes and could not be built. Key changes:

- **Build-Depends**: add `libsystemd-dev`, `dh-python`, `chrpath`; remove `python3-pywbem` (no longer in Debian)
- **Package split**: replace `libstoragemgmtd` / `libstoragemgmt-tools` / `libstoragemgmt-plugins` with per-plugin packages (`libstoragemgmt-{targetd,megaraid,hpsa,arcconf,nfs,local}-plugin`) plus `libstoragemgmt-udev`
- **Configure flags**: `--without-smispy` (python3-pywbem unavailable) and `--without-ledmon` (libled not packaged in Debian) — both made optional by PR #560
- **udev files**: install `90-scsi-ua.rules` and `scan-scsi-target` manually in `override_dh_auto_install` since they are `noinst` in `Makefile.am`
- **debhelper-compat 13**, Standards-Version 4.6.2
- Empty `debian/patches/series`: all fixes are commits in the tree

## Test plan

- [ ] `dpkg-buildpackage -us -uc -b` succeeds on Debian 12 (bookworm)
- [ ] `dpkg-buildpackage -us -uc -b` succeeds on Debian 13 (trixie)
- [ ] `dpkg-buildpackage -us -uc -b` succeeds on Ubuntu 24.04
- [ ] `lsmcli list --type systems -u sim://` returns expected output after install
- [ ] `libstoragemgmt-udev` installs rules to `/usr/lib/udev/rules.d/90-scsi-ua.rules`

## Summary by CodeRabbit

* **New Features**
  * Reworked Debian packaging into dedicated components, including dedicated controller/plugin packages and added udev rules plus a SCSI scan helper.

* **Bug Fixes**
  * Updated dependency wiring and install manifests to ensure each package ships the intended runtime, development, and plugin files.
  * Improved systemd directory detection for more consistent defaults.

* **Chores**
  * Refreshed Debian metadata (changelog/compat/watch), updated licensing/copyright, and removed obsolete packaging entries/scripts.

CI-kick


tasleson:
@brianatpurestorage 

Thanks for this deb package updates.  I'm not a debian packaging person, so I used Claude Code to assist with a review of this PR.

  ### Issues that should be looked at/fixed

  **1. Circular dependency between `libstoragemgmt1` and `python3-libstoragemgmt`**

  `libstoragemgmt1` explicitly Depends on `python3-libstoragemgmt (= ${binary:Version})`, and `python3-libstoragemgmt` explicitly Depends on `libstoragemgmt1 (= ${binary:Version})`. dpkg can handle this, but lintian will flag it and Debian policy discourages it. The RPM spec avoids this: the main package requires python3, but python3 only has an implicit dependency back on the libs (via the .so linkage).

The root cause is that `libstoragemgmt1` maybe bundles too much (see next point)?  If splitting isn't desired, the explicit reverse dependency in `python3-libstoragemgmt` on `libstoragemgmt1` should be removed,  `${shlibs:Depends}` already handles it via `_clib.so` linking against `libstoragemgmt.so.1`.

  **2. `libstoragemgmt1` bundles daemon, CLI, config, systemd units, and the simulator plugin**

  In Debian convention, `lib<name>N` should contain only the shared library (`.so.N.*`). The current `libstoragemgmt1.install` includes `lsmcli`, `lsmd`, `simc_lsmplugin`, systemd service/tmpfiles/sysusers configs, bash completion, man pages, and `lsmd.conf`. This should ideally be split: `libstoragemgmt1` for the shared lib, and a separate package (e.g. `libstoragemgmt-daemon` or just `libstoragemgmt`) for the tools/daemon. This is what the RPM spec does with `%files` (tools+daemon) vs `%files -n %{name}-libs` (just the .so).

  If a split is too much churn for this PR, at minimum rename the package to something that doesn't look like a pure library package (e.g. `libstoragemgmt` instead of `libstoragemgmt1`), or document the bundling decision.
  
 Documentation on shared libs, https://www.debian.org/doc/debian-policy/ch-sharedlibs.html

  **3. `Multi-Arch: same` is incorrect on `libstoragemgmt1`**

  The package has files in `/usr/bin/` (lsmcli, lsmd, simc_lsmplugin), `/etc/`, `/usr/lib/systemd/`, etc. These are arch-independent paths that would conflict if multi-arch co-installed. `Multi-Arch: same` should be removed from this package. It would be correct only if the package contained just the shared library in the multiarch lib path.

  **4. `Multi-Arch: same` is incorrect on `python3-libstoragemgmt`**

Python modules in `/usr/lib/python3*/dist-packages/` are not multiarch-safe — the `.py` files are identical across architectures. This should be removed or changed to `Multi-Arch: foreign`.

  **5. `Multi-Arch: same` on `libstoragemgmt-dev` is questionable**

 Headers in `/usr/include/libstoragemgmt/` and man pages in `/usr/share/man/man3/` are arch-independent and would conflict. The `.so` symlink and `.pc` file are arch-specific. For -dev packages, `Multi-Arch: same` requires the headers to be identical across arches (which they are here), but the man pages in `/usr/share/` would conflict. Either remove `Multi-Arch: same` or ensure the paths are correct for multiarch.

  **6. Redundant `./autogen.sh` in `override_dh_auto_configure`**

  With debhelper-compat 13, `dh_autoreconf` runs automatically before `dh_auto_configure` and already calls `autoreconf -fi`. The explicit `./autogen.sh` call makes autoreconf run twice. Remove it — `dh_autoreconf` handles this.

  ### Issues possibly worth addressing

  **7. Missing `Files: debian/*` in `debian/copyright`**

  The new copyright file has detailed stanzas for upstream files but no `Files: debian/*` section attributing the packaging copyright to Brian Bunker. Debian policy expects this.

  **8. Missing `Vcs-Git` / `Vcs-Browser` fields**

 The old control had these (pointing to stale alioth URLs, so correctly removed). They should be updated to the GitHub URLs:

  Vcs-Git: https://github.com/libstorage/libstoragemgmt.git
  Vcs-Browser: https://github.com/libstorage/libstoragemgmt

  **9. `nfs-kernel-server` as a hard dependency for `libstoragemgmt-nfs-plugin`**

Installing the NFS plugin currently forces installing the full NFS kernel server. The plugin only needs the server when actively managing exports, not just to be present. This should be `Recommends` or `Suggests`.

  **10. `DH_VERBOSE = 1` enabled by default**

Produces very noisy build logs. Fine for development but should be `DH_VERBOSE = 0` (or commented out) for a released package.

  ### Minor / Nitpicks

  **11. `usr/lib*/lsm.d/` glob is unconventional**

  In `python3-libstoragemgmt.install`, the glob `usr/lib*/lsm.d/` works to match `usr/libexec/lsm.d/` (since `libexec` starts with `lib`), but it's subtle. Using the explicit `usr/libexec/lsm.d/` would be clearer.

  **12. Copyright year `2011-2023`**

  Slightly stale for 2026, update?

tasleson:
@brianatpurestorage The one failing CI test is fine, I need to dig into that one, but it obviously has nothing to do with this.

Otherwise, I think the only small nit is that the changelog day-of-week is wrong,  June 17, 2026 is a Wednesday, not a Tuesday.

  ```diff
  - -- Brian Bunker <brian@purestorage.com>  Tue, 17 Jun 2026 00:00:00 +0000
  + -- Brian Bunker <brian@purestorage.com>  Wed, 17 Jun 2026 00:00:00 +0000

  dpkg-parsechangelog will reject an incorrect day-of-week.
  ```

Thanks!

Also, we can certainly roll a new release, before you create any debian packages, if that helps you out.

tasleson:
@brianatpurestorage Let me know when all the debian package builds work, and I'll merge this, thanks!

brianatpurestorage:
Build verified on Debian 12 (bookworm) arm64 — all packages build cleanly: `libstoragemgmt1`, `libstoragemgmt`, `libstoragemgmt-dev`, `python3-libstoragemgmt`, and all 7 plugin/udev packages. Ready to merge.

tasleson:
> Build verified on Debian 12 (bookworm) arm64 — all packages build cleanly: `libstoragemgmt1`, `libstoragemgmt`, `libstoragemgmt-dev`, `python3-libstoragemgmt`, and all 7 plugin/udev packages. Ready to merge.

Thanks!

PR #563, opened 2026-06-29
ci: add Debian bookworm CI and fix !HAVE_LEDMON argument validation

brianatpurestorage:
## Summary

- Adds Debian bookworm to the CircleCI matrix alongside the existing Fedora job
- Fixes `docker_ci_test.sh` configure flags for Debian (`--without-mem-leak-test --without-smispy --without-ledmon`)
- Adds `libsystemd-dev` and `pkg-config` to `deb_dependency` for the build container
- Runs `make deb` at the end of the Debian CI job to validate package builds
- Fixes the `!HAVE_LEDMON` stub in `lsm_local_disk_led_status_get`: the stub was silently discarding NULL arguments and returning `LSM_ERR_NO_SUPPORT`; it now validates arguments (`LSM_ERR_INVALID_ARGUMENT`) and checks disk existence (`LSM_ERR_NOT_FOUND_DISK`), matching what the test suite expects

## Test plan

- [ ] CircleCI Debian job passes `make check` and `make deb`
- [ ] CircleCI Fedora job continues to pass (no regression)
- [ ] `test_local_disk_led_status_get` passes on a build without ledmon support

## Summary by CodeRabbit

* **New Features**
  * Added Debian-based CI coverage alongside the existing Fedora job.
  * Expanded Debian packaging support with additional required build dependencies.

* **Bug Fixes**
  * Improved disk LED status handling so missing disks now report a clear “not found” error instead of only a generic unsupported result.
  * Set sensible default LED status values when the feature is unavailable.

* **Chores**
  * Updated Debian build and CI configuration for the new platform-specific setup.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,21 @@ jobs:
          command: ./test/docker_ci_test.sh
          no_output_timeout: 20m
 
+  debian:
+    docker:
+      - image: debian:bookworm
+    steps:
+      - run:
+         command: apt-get update -q && apt-get install -y git
+         no_output_timeout: 5m
+      - checkout
+      - run:
+         command: ./test/docker_ci_test.sh
+         no_output_timeout: 20m
+
 workflows:
   version: 2
   workflow:
     jobs:
     - fedora
+    - debian
--- a/c_binding/lsm_local_disk.cpp
+++ b/c_binding/lsm_local_disk.cpp
@@ -1360,8 +1360,29 @@ out:
     led_free(ctx);
     return rc;
 #else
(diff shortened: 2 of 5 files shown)


brianatpurestorage:
Good catch — fixed in the latest push. Moved the `*lsm_err = NULL` clear to before `_check_null_ptr()` (guarded by `lsm_err != NULL`), and updated the `out:` block to reset `*led_status` and set `*lsm_err` using the same null-guard pattern as the `HAVE_LEDMON` branch.

tasleson:
Thanks @brianatpurestorage

Issue #564, opened 2026-07-24
MegaRAID storcli output:'Model Number' is missing


tasleson:
The libstoragemgmt megaraid plugin expects a `Model Number` key in `storcli` disk output, but storcli
 version 007.3703.0000.0000 doesn't include it. This caused lsmcli list --type='DISKS' to fail with exit code 4 for both
 megaraid:// and megaraid+ssl:// URIs.

  Cascade: No disks enumerable → 0 free disks → all RAID types skipped (RAID0, RAID5, RAID6, RAID50, RAID60).

  Fix needed: The libstoragemgmt megaraid plugin needs to handle the missing 'Model Number' key gracefully (make it
  optional or map it to the new field name in this storcli version).

tasleson:
Root cause: The MegaRAID 9560-16i controller on korg has NVMe drives behind it. For NVMe drives, storcli reports the model as "Model" in the device attributes section, while for SAS/SATA drives it uses "Model Number". The plugin at megaraid.py:605 was unconditionally accessing disk_show_attr_dict['Model Number'], which throws a KeyError for NVMe drives, caught by the _handle_errors decorator and reported as PLUGIN_BUG.   Our CI test hardware does not have NVMe drives.

tasleson:
proposed fix

```diff
diff --git a/plugin/megaraid_plugin/megaraid.py b/plugin/megaraid_plugin/megaraid.py
index a2755d55..2b8ae070 100644
--- a/plugin/megaraid_plugin/megaraid.py
+++ b/plugin/megaraid_plugin/megaraid.py
@@ -599,10 +599,13 @@ def disks(self,
                     'Drive %s State' % mega_disk_path]
 
                 disk_id = disk_show_attr_dict['SN'].strip()
+                disk_model = disk_show_attr_dict.get(
+                    'Model Number',
+                    disk_show_attr_dict.get('Model', ''))
                 disk_name = "Disk %s %s %s" % (
                     disk_show_basic_dict['DID'],
                     disk_show_attr_dict['Manufacturer Id'].strip(),
-                    disk_show_attr_dict['Model Number'])
+                    disk_model)
                 disk_type = _disk_type_of(disk_show_basic_dict)
                 blk_size = size_human_2_size_bytes(
                     disk_show_basic_dict['SeSz'])
```

PR #566, opened 2026-08-05
megaraid: fix NVMe JSON key

tasleson:
When a megaraid controller has NVMe devices, it appears that "Model Number" becomes "Model".  I checked the last 8 versions of storecli and they had the same behavior, so it's been this way for a long time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MegaRAID disk name detection across supported StorCLI output formats.
  * Added fallback handling when the primary model information is empty, unavailable, or missing.
  * Prevented incomplete model data from producing invalid disk names.
* **Tests**
  * Added coverage for model selection, fallback behavior, and missing disk attributes.
  * Enabled standalone MegaRAID validation without requiring the full daemon environment.

### DIFF
--- a/deb_dependency
+++ b/deb_dependency
@@ -15,6 +15,7 @@ libudev-dev
 libsystemd-dev
 pkg-config
 python3-dev
+python3-pytest
 perl
 valgrind
 chrpath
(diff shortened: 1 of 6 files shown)


tasleson:
Resolves: https://github.com/libstorage/libstoragemgmt/issues/564

